PR #2, opened 2015-07-05
fix false literal typo

ryanatkn:


ryanatkn:
Looking at this again, the length comparison in `argsEquals` should never return false unless you mutate the selector array passed to `createSelector`. I think you can just remove that check - don't see why you would want to ever change the dependencies of a result function at runtime.

ellbee:
@faassen I agree with ryanatkn, the length comparison check does look redundant here. What do you think?

faassen:
Ah, lost track of this one. Yes, I took a look at it and I agree. It'd be only be needed if this function was to be reused in other cases, but it's not so we're fine.

faassen:
This reminds me we should install some kind of coverage tool for the tests so we can find the original False spelling issue (my Python reflexes misleading me) more quickly. Does anyone know one we can integrate? I'll create an issue, see #5.

faassen:
If we change this one, how does this interact with hot loading, though? When you are doing hot loading it's possible for someone to change the signature of a selector. But would hot loading cause everything to be reloaded so we're okay, or would we have trouble with the cache having the old number of arguments?

ellbee:
If the signature of a selector changes it means it has been edited, and therefore reloaded. So I don't think there should be a problem.

Issue #1, opened 2015-07-01
keypaths


faassen:
We could implement key path arrays as an argument to `createSelector` as a shortcut, i.e instead of this:

```
createSelector([state => state.a.b], value => value);
```

we could allow this:

```
createSelector([['a', 'b']], value=>value);
```

Or alternatively a string spelling:

```
createSelector(['a.b'], value=value);
```

to get an object.

We should make how to interpret a key path pluggable in `createSelectorCreator`. This way application authors can deal with special cases where state is a ES6 Map or an Immutable.js object, or whatever else.

We could even interpret anything that's not a function as a "keypath" and pass it along to the key path resolver function, in which case using an array or a dotted name is up to the application author.

ellbee:
Our first issue! I am going to close it out now because no-one has asked for it, so I don't think it is going to get implemented.

PR #4, opened 2015-07-06
Suggestion: add default binding for result function

speedskater:
As the base case for selectors is often to only extract a field of the current state it is quite handy to bind the result function to the identity function.

The question is whether it makes sense binding the identity function to the result function or to suggest that the user should specify default selectors as plain functions (e.g. in the README const shoptItemSelector = state => state.shop.items)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,10 @@ import React from 'react';
 import { createSelector } from 'reselect';
 import { connect } from 'redux/react';
 
+const shopItemSelector = createSelector(state => state.shop.items);
+
 const subtotalSelector = createSelector(
-  [state => state.shop.items],
+  shopItemSelector,
   items => items.reduce((acc, item) => acc + item.value, 0)
 );
 
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,7 @@
+const identitySelector = state => state;
+
 export function createSelectorCreator(valueEquals) {
-    return (selectors, resultFunc) => {
+    return (selectors, resultFunc = identitySelector) => {
         if (!Array.isArray(selectors)) {
             selectors = [selectors];
         }
--- a/test/test_selector.js
+++ b/test/test_selector.js
@@ -13,6 +13,10 @@ suite('selector', function() {
             [state => state.a, state => state.b], (a, b) => a + b);
         assert.equal(selector({a: 1, b: 2}), 3);
     });
+    test("default result function", function() {
+        const selector = createSelector([state => state.a]);
+        assert.equal(selector({a: 1}), 1);
+    });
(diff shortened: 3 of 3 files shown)


faassen:
@ellbee and I had a discussion about this at the hackathon. If I recall correctly we had two issues with the approach:

* is this actually useful in practice? You refactor a selector out of the first example, but is such a stand-alone selector useful or just more verbose?

* the identity function only works for selectors that take only a single arguments. If a selector combines multiple other selectors this goes wrong. At the very least we need to define some behavior (exception?) and a test case that tests this.

@ellbee, please do add to this; I'm actually a total noob at selectors myself. :)

ellbee:
I guess the stand-alone selector is useful for abstracting away the shape of the redux store?

Like if you had the following:

```js
store :  {
  inner1: {
    inner2: {
    }
  }
}

const selectInner2 = createSelector(state => state.inner1.inner2);
```
You could use selectInner2 in multiple components to access inner2. Then if, for example, you wanted to start using Immutable.js, you would only need to change the selector.

```js
Immutable.fromJS(
  store: {
    inner1: {
      inner2: {
      }
    }
  }
});

const selectInner2 = createSelector(state => state.getIn(['inner1', 'inner2']));
```

faassen:
Yes, I can see there is an argument for abstraction -- multiple components can use the same selector to access a value, so that if you change the structure of the store, you don't need to modify multiple places.

This still leaves the whole issue of the identity function only working when a single argument is involved. Since we have a different signature (only a single selector argument, no second argument), this implies to me a different function. What about a `createAccessor` that takes only a single argument for this specific purpose, instead of overloading the behavior of `createSelector`? We can still layer it over `createSelector`, though we need to think a bit about how to make it work with `createSelectorCreator`.

ellbee:
Am I crazy, or is
```js
const selectInner2 = createSelector(state => state.inner1.inner2);
```
actually exactly the same thing as
```js
const selectInner2 = state => state.inner1.inner2;
```
???

speedskater:
@ellbee Sounds good for me. I can create a corresponding pull request regarding the README tomorrow morning.
@faassen I like your suggestion for createAccessor. This would solve https://github.com/faassen/reselect/issues/1 as well. The question again is whether such a convenience function is needed. Although such keypaths might safe some characters it will definitively be computationally more expensive than plain functions.

faassen:
+1 on adding more documentation about making them plain functions. 

If we decide to go for #1 we can always create createAccessor later as it'd be needed then.

ellbee:
A pull request would be great, thanks. I have added a bit to the API Documentation section, maybe you can add an example using a simple access selector function there too?

PR #6, opened 2015-07-07
Extended Examples for Simple Selector

speedskater:
As discussed in https://github.com/faassen/reselect/pull/4#issuecomment-119001714 I revised the Examples to point out the difference between simple and combined selectors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ Simple "selector" library for Redux inspired by getters in [nuclear.js](https://
 
 ## Example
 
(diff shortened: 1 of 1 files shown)


ellbee:
Ok, I think this is very good. Thanks @speedskater!

ellbee:
@faassen Do you have any comments on this? If you are ok with it I'll pull it in.

speedskater:
Thanks for merging.

PR #9, opened 2015-07-14
Minor perf improvements

### LINKED ISSUE #8: don't check for argument length comparison
Just so we don't lose track of this issue that came up on #2. It would be a small performance optimization.

We could do away with this check entirely. Pull requests are welcome.

@ryanatkn @ellbee

HeyImAlex:
First one should close #8. Second one just makes sense to me: we create params as an array, then spread it to memoize only to have memoize coerce it back to an array. Compile with babel and check out how much extraneous code it generates.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,8 +38,5 @@ function memoize(func, valueEquals) {
 }
 
 function argsEquals(a, b, valueEquals) {
-    if (a.length != b.length) {
-        return false;
-    }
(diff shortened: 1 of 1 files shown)


faassen:
Okay, looks good. One possible objection is that now the signature of the memoized function is different from the original function, but since the whole memoization is happening in private functions anyway, that's really an implementation detail.

Issue #7, opened 2015-07-10
Add a way to shortcut evaluation


HeyImAlex:
Just an idea for dealing with situations where a selector function should be conditional.

Here's a contrived example: Imagine you have a huge list and a toggle button to do some complex filter calculations on the list. Your base selectors look like this.

```javascript
const hugeList$ = state => state.hugeList;
const shouldFilter$ = state => state.shouldFilter;
```

One way to handle this is by making a `filteredList$` selector that depends on `hugeList$`.
 
```javascript
// Method #1

const filteredList$ = createSelector(
  [hugeList$],
  (hugeList) => {
    /* do expensive computation here */
    return filteredList;
  }
);

const finalList$ = createSelector(
    shouldFilter$, filteredList$, hugeList$],
    (shouldFilter, filteredList, hugeList) => {
        return shouldFilter ? filteredList : hugeList;
    }
);
```

The issue is that `filteredList` is computed even if `shouldFilter` is false. If `hugeList` changed very often but `shouldFilter` was usually false, you'd be doing a lot of unnecessary re-calculation.

To avoid that you could move the `shouldFilter` check into `filteredList$`;

```javascript
// Method #2

const filteredList$ = createSelector(
  [shouldFilter$, hugeList$],
  (shouldFilter, hugeList) => {
    if (!shouldFilter) { return null; }
    /* do expensive computation here */
    return filteredList;
  }
);

const finalList$ = createSelector(
    [filteredList$, hugeList$],
    (filteredList, hugeList) => {
        return filteredList === null ? hugeList : filteredList;
    }
);
```

But now the problem is that you recompute every time `shouldFilter` is toggled. If someone repeatedly clicks the toggle button the UI is going to start lagging hard.

The crux of the issue is that we have no way to conditionally execute a selector. We want to compute a value _under certain circumstances only_, but using memoize we can't shortcut the function without expiring our cached value.

Some potential solutions:

- A Symbol that can be returned to shortcut execution directly. The wrapping function would check the return value and return the cached value if it was the symbol. In the example you could just take the second method and return the special symbol instead of null.
- Something like `shouldSelectorUpdate` that passes the last _used_ dependency values and the current dependency values and lets the user decide if they wanted to recompute.

The only issue I see now is the need for a default value. As this situation only comes up times that you don't actually want to use the value, it seems safe to just default to null and let the user override if they need to.

ellbee:
[...]

PR #21, opened 2015-08-06
make memoize exportable

ellbee:
Addresses #7

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,7 +3,7 @@ export function createSelectorCreator(valueEquals) {
         if (!Array.isArray(selectors)) {
             selectors = [selectors];
         }
-        const memoizedResultFunc = memoize(resultFunc, valueEquals);
+        const memoizedResultFunc = internalMemoize(resultFunc, valueEquals);
         return state => {
             const params = selectors.map(selector => selector(state));
             return memoizedResultFunc(params);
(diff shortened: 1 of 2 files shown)


faassen:
I haven't followed along various discussions (I need to catch up but work is so busy lately) so I may be talking nonsense. I vaguely recall thinking at some point: it's a good thing memoize is internal as it's not really a general memoize function. Does that objection make any sense?

ellbee:
Yeah, there are two things about this function which the user might not expect. 

* The cache size is fixed to 1
* We don't serialise/hash the value of the arguments when they are objects (i.e. if you deeply mutate an argument we will always get a cache hit due to reference equality).

Possible solutions:

* Documentation
* Rename exported memoize function
* Make memoize less surprising by implementing a variable cache size and providing the option for value equality for objects, maybe by using a third party memoize function (lodash?). If we did this I guess there is no reason to export the function at all because I think the original problem could be solved by changing the cache and equality settings.

ellbee:
@HeyImAlex, @speedskater Any thoughts on making memoize more configurable instead of exporting it?

speedskater:
@ellbee sorry for the delay but i was a bit busy the last days.

I like your idea making the memoization strategy configurable as a configuration option for creating selectors (there it would be able to use lodash memoization, database based ones or other fancy strategies). 

As far as i understand, in our internal use case the default memoization functions should  be a function which uses cache size 1 and does reference equality checking. The reason is that a common memoization function like the one provided by lodash might store all previous call results, which can result in a significantly increased memory footprint for the whole application.

Maybe for our internal memoization function the name memoize is missleading. So maybe it might be less confusing if we name the function buffer or createCached or something totally different. Than it might be less confusing in the exported case. Or do you know of any library which exports the same kind of memoization/buffer function?
What do you think?

banderson:
re: configuration, the one thing I see as obviously useful would be the cache size. Having it set to 1 was a surprise to me. Keeping the reference equality constraint is a good thing I think, and expected since it's advertised as using with redux which requires immutability.

ellbee:
@banderson Yes, I agree about the cache size being surprising, that is definitely something I would like to be configurable. The downside is that I think allowing a cache size bigger than 1 would mean that straight reference equality is not going to work with objects. Redux's immutability means the object reference (and therefore the key into the cache) is going to be different each time. Even when an objects value is the same as some previous object in the cache it won't produce a cache hit. We'd need to hash or JSON.stringify the object to get a value we could use a cache key.

ellbee:
@speedskater I pretty much agree. I'll have a think about whether there is a better name for memoize.

banderson:
@ellbee I see the issue there. That could still be useful for those providing a custom equality function, however. No urgent need here, just my 2 cents :)

ellbee:
Probably superseded by #32

Issue #27, opened 2015-08-17
Allow variadic dependencies as well as an array


ccapndave:
As well as the current system of `createSelector([a, b, c], (a, b, c) => ...)` it would be great if it would also be possible to do `createSelector(a, b, c, (a, b, c) => ...)`

This would allow us to implement robust and strictly typed Typescript definitions on `createSelector`.

ellbee:
Oh, very interesting. The issue is that all values in array have to have the same type?

ccapndave:
Exactly (or you can specify a union type, but not the order they come in).  If they are separate arguments they can be individually typed.  This is how Rx gets such good typing in Typescript for functions such as `comineLatest`.

ellbee:
Ok, thanks. I don't see why we shouldn't do this. @faassen?

speedskater:
@ccapndave @ellbee Sounds interesting to me. 
I think it has two more advantages:
* it deals with a common mistake that array brackets are ommited.
* ommiting the array brackets makes it imho more readable.

ellbee:
Variadic dependencies added to the 1.0.0 release

Issue #20, opened 2015-08-06
Can't pass props in selectors


vasa-chi:
`react-redux` supports [passing props to select](https://github.com/gaearon/react-redux/blob/master/src/components/createConnectDecorator.js#L18). Currently reselect only passes state. Is it possible to add props to selector functions?

ellbee:
I'm not sure that the new react-redux api for the connect decorator is going to support this. [Check out this PR](https://github.com/gaearon/react-redux/pull/16)

vasa-chi:
Huh. That would be unfortunate... Maybe @gaearon can clarify?

vasa-chi:
Seems that there will be an additional argument for `@connect` decorator for using props to further filter the state. So I guess that yes, this api will get deprecated.

ellbee:
Confirmed as gone in the [release notes for react-redux v.0.5.0](https://github.com/gaearon/react-redux/releases/tag/v0.5.0)

scottcorgan:
It looks like this added back in react-redux. Can we get it back in this package?

ellbee:
Yep, I guess we should look into this now

scottcorgan:
Any time frame on this? Attempted to jump in on a fork, but didn't know if changes were coming soon for a 1.0.

ellbee:
Funnily enough, I was just looking at it. Did you have a plan? I'm wondering whether the dumb default memoize and its reference equality is going to work with the props. I don't recall how the props are handled inside react. I guess if the prop object is mutative and has a stable reference for the lifetime of the component *or* it is recreated entirely on each render then we might have problems?

I think that the API might change slightly in the next version, possibly allowing the user to [specify a memoization function in createSelectorCreator](7) and allowing [variadic depencies](27). I just started exploring this in my fork [here](https://github.com/ellbee/reselect.git) if you'd like to have a look.

gaearon:
>I guess if the prop object is mutative and has a stable reference for the lifetime of the component or it is recreated entirely on each render then we might have problems?

Props object is new every time. It's considered immutable.

faassen:
Okay, I just finally started using reselect in a project and ran into this one. So basically we want to parameterize the select expression with an additional object that should be passed down the select chain, right? And we should be able to express when we consider this object unchanged, so that we can make use of caching.

ellbee:
Take a look at #35. We've had a bit of a discussion about this there

ellbee:
Address in the 1.0.0 release

PR #33, opened 2015-08-22
WIP 0.0.3 (Configurable memoize / variadic dependencies / pass props to selector)

ellbee:
Work in progress experiments to address issues #7 , #20 , #27. The API section of the README has some examples.

There is one breaking change. `createSelectorCreator` now takes a memoize function instead of a valueEquals function.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,12 @@
+{
+  "extends": "eslint-config-airbnb",
+  "env": {
+    "browser": true,
+    "mocha": true,
+    "node": true
+  },
+  "rules": {
+    "block-scoped-var": 0,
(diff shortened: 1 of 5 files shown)


gaearon:
If it has a breaking change, it better be `0.1.0`, not `0.0.3`.
Otherwise you'll break any consumers that have `^0.0.2` in `package.json`.

gaearon:
Oh wait, I'm wrong, you're still `0.0.x`.
Anyway, I suggest you jump to 1.0 soon :-). People are using it in production.

ellbee:
Yes, good point. I'll rename this and make it the 1.0.0 branch.

PR #35, opened 2015-08-25
WIP 1.0.0 (Configurable memoize / variadic dependencies / pass props to selector)

ellbee:
Work in progress experiments to address issues #7 , #20 , #27. The API section of the README has some examples.

There is one breaking change. createSelectorCreator now takes a memoize function instead of a valueEquals function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,9 @@ export default connect(visibleTodosSelector)(App);
 
 ### Accessing React Props in Selectors
 
-The following code shows a modification to `index.js` where a `maxTodos` prop is being passed to the `App` component that specifies the maximum number of Todos to be displayed at any one time:
+So far our selectors have only been receiving the Redux store as input. It is also possible to pass the props of a component wrapped by `connect` into a selector.
+
+Consider the following example:
 
 #### `index.js`
 
@@ -274,7 +276,7 @@ React.render(
 
 ```
 
-When a selector is connected to a component with `connect`, the component props are passed as the second argument to the selector. We will use this fact to access the `maxTodos` prop from within `visibleTodosSelector`. First, we create a new input-selector named `maxTodosSelector` which gets `maxTodos` from its props argument (and ignores its state argument). Then we add `maxTodosSelector` as an input selector to `visibleTodosSelector`:
+The prop `maxTodos` has been added to the `App` component. We would like to access this prop in `visibleTodosSelector` so we change `selectors/todoSelectors.js` to the following:
 
 #### `selectors/todoSelectors.js`
 
@@ -295,7 +297,6 @@ function selectTodos(todos, filter) {
 
 const visibilityFilterSelector = state => state.visibilityFilter;
 const todosSelector = state => state.todos;
-// accessing props via the second argument
 const maxTodosSelector = (_, props) => props.maxTodos;
 
 export const visibleTodosSelector = createSelector(
@@ -311,6 +312,8 @@ export const visibleTodosSelector = createSelector(
   }
 );
 ```
+
+When a selector is connected to a component with `connect`, the component props are passed as the second argument to the selector. In the example above, we added a new input-selector named `maxTodosSelector` which gets `maxTodos` from its props argument (and ignores its state argument). `maxTodosSelector` was then added as an input selector to `visibleTodosSelector`, making `maxTodos` available to the selectors result function.
 
 ## API
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ export const totalSelector = createSelector(
 
 ### Motivation for Memoized Selectors
 
-> The examples in this section are based on the [Redux Todos List example](https://github.com/docs/basics/UsageWithReact.md).
+> The examples in this section are based on the [Redux Todos List example](http://rackt.github.io/redux/docs/basics/UsageWithReact.html).
 
 Consider the following code:
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,9 @@ export const totalSelector = createSelector(
 
 ### Motivation for Memoized Selectors
 
-Here is `App.js` from the [Redux Todos List example](https://github.com/docs/basics/UsageWithReact.md):
+> The examples in this section are based on the [Redux Todos List example](https://github.com/docs/basics/UsageWithReact.md).
+
+Consider the following code:
 
 #### `containers/App.js`
 
@@ -165,7 +167,6 @@ function selectTodos(todos, filter) {
  * of the store in cases where no calculations are needed 
  * and memoization wouldn't provide any benefits.
  */
-
 const visibilityFilterSelector = state => state.visibilityFilter;
 const todosSelector = state => state.todos;
 
@@ -176,7 +177,6 @@ const todosSelector = state => state.todos;
  * Hence, these selectors are only recomputed when the value of their input selectors change. 
  * If none of the input selectors return a new value, the previously computed value is returned.
  */
- 
 export const visibleTodosSelector = createSelector(
   visibilityFilterSelector,
   todosSelector,
@@ -224,7 +224,6 @@ import Footer from '../components/Footer';
  * Import the selector defined in ../selectors/todoSelectors.js.
  * This allows you to separate your components from the structure of your stores.
  */
-
 import { visibleTodosSelector } from '../selectors/todoSelectors';
 
 class App extends Component {
@@ -265,7 +264,7 @@ App.propTypes = {
 };
 
 /*
- * Bind the visibleTodosSelector to the App component.
+ * Connet visibleTodosSelector to the App component.
  * The keys of the selector result are available on the props object for App.
  * In our example there is the 'visibleTodos' key which is bound to this.props.visibleTodos
  */
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ const todosSelector = state => state.todos;
 /* 
  * Definition of combined selector. 
  * In visibleTodosSelector, input selectors are combined to derive new information. 
- * To prevent expensive recalculation of the input-selectors memoization is applied. 
+ * To prevent expensive recalculation of the input selectors memoization is applied. 
  * Hence, these selectors are only recomputed when the value of their input selectors change. 
  * If none of the input selectors return a new value, the previously computed value is returned.
  */
--- a/README.md
+++ b/README.md
@@ -159,9 +159,24 @@ function selectTodos(todos, filter) {
   }
 }
 
+/*
+ * Definition of input selectors. 
+ * Input selectors should be used to abstract away the structure
+ * of the store in cases where no calculations are needed 
+ * and memoization wouldn't provide any benefits.
+ */
+
 const visibilityFilterSelector = state => state.visibilityFilter;
 const todosSelector = state => state.todos;
 
+/* 
+ * Definition of combined selector. 
+ * In visibleTodosSelector, input selectors are combined to derive new information. 
+ * To prevent expensive recalculation of the input-selectors memoization is applied. 
+ * Hence, these selectors are only recomputed when the value of their input selectors change. 
+ * If none of the input selectors return a new value, the previously computed value is returned.
+ */
+ 
 export const visibleTodosSelector = createSelector(
   visibilityFilterSelector,
   todosSelector,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,40 +57,42 @@ const taxPercentSelector = state => state.shop.taxPercent;
...
 
 const subtotalSelector = createSelector(
-  [expensiveItemSelectorFactory(200)],
+  expensiveItemSelectorFactory(200),
   items => items.reduce((acc, item) => acc + item.value, 0)
 );
 ```
 
-### Selector Usage
+### Using Selectors with React-redux
```
gaearon: Let's capitalize as React Redux (like in docs)

README.md
```diff
@@ -121,34 +123,91 @@ export default Total;
 
 ### createSelector([inputSelectors], resultFn)
```
gaearon: Should be `createSelector(...selectors, resultFn)`

gaearon: Oh, I see, you allow an array too..

README.md
ellbee: Agreed

README.md
ellbee: Yeah, `...inputSelectors` was added for this new version, we only allow the array in 0.0.2. I'm going to add the `...inputSelectors` signature here too.

BTW, thanks for taking a look at this!

ellbee:
@speedskater @faassen @HeyImAlex (and anyone else reading!)

If anyone has some spare time on their hands it would be great if I get could get some opinions on this PR.

* There is a breaking change in createSelectorCreator, are we happy to make a breaking change?
* It was suggested by @gaearon that we should consider moving to 1.0.0, does anyone have any objections to doing so? (I don't)
* I am currently rewriting the readme, but I made some quick changes to it in this PR to give a couple of example of the new API

gaearon:
>I am currently rewriting the readme, but I made some quick changes to it in this PR to give a couple of example of the new API

I think everybody's waiting for “how to use this with React Redux `ownProps` parameter”

ellbee:
The last couple of examples for `createSelector` should show roughly how it works at the moment. I am holding off on writing the docs for this as I'm not sure if it is going to stick, let me know if further explanation is needed. I have added a note to the `defaultMemoizeFunc` section with some thoughts about why I'm not happy with it.

README.md
```diff
@@ -119,36 +121,106 @@ export default Total;
...
     state => state.values.value2
   ],
   (value1, value2) => value1 + value2
 );
 
-// it is not necessary to wrap a single input selector in an array
-const totalSelector = createSelector(
-  state => state.shop.items,
-  items => items.reduce((acc, item) => acc + item.value, 0)
+// A selector is also passed props as the last parameter into the resultFunc
```
speedskater: Can this result in problems, as the selector suddenly depends on all properties.
Wouldn't it be better to have the following signature

createSelector(
  state, props => state.values.value1,
  state, props => state.values.value2,
  state, props => props.value3,
  value1, value2, value2 => value1, value2, value3
)

README.md
```diff
@@ -119,36 +121,106 @@ export default Total;
...
+```
+
+`defaultMemoizeFunc` does not look at the props argument when determining if the arguments have changed. This is because by default `defaultMemoize` uses `defaultValueEquals`, and as`defaultValueEquals` is a reference equality check and the reference of the props will be different each render the memoization will not work. 
+
+(NOTE: I am not happy about this, I think this is going to surprise people. I am considering adding an option that specifies whether the props should be used for memoization as part of the configuration options for people who aren't using the reference equality of `defaultValueEquals`. To make things worse, when using a third party memoize like `lodash.memoize` with `createSelectorCreator`, the memoize function **will** take the props into account for memoization. I think it is likely that if you are using a third party memoize your hashing function will be doing some kind of deepEquals so the selector will only update when there is actually a new value somewhere in the props. However, this could still be a problem if the props are changing in a manner such that they resolve to a new never-before-seen hash on every render. Ideas are welcome here.)
```
speedskater: Could the suggestion of line 145 mitigate this problem.  
We could go even further to allow arbitrary number of arguments to the selection function and the selectors determining the arguments for the memoized function have to deal with the selection of the correct properties. Therefore only in case the actual input properties changes does the selected value changes. 
An alternative approach would be to use a compound object of the form ({ props, state }). The latter approach wouldn't force us to change the signature of the select function at all. It would furthermore allow us the to provide convenience selectors like selectWithProp('state', stateSelector), selectWithProp('props', propertySelector).....

I think this approach could enable reselect to achieve the same semantics as if mapStateToProps would be changed into a curried version props => state => selectionFunction

@ellbee, @faassen, @gaearon  What do you think?

ellbee:
@speedskater Thanks for the feedback. Inspired by your first suggestion, I simply changed `memoizedResultFunc` so that it doesn't take the props argument. This gets rid of the special casing for `defaultMemoize`, so I think it is an improvement.

speedskater:
@ellbee  Thanks. Do you really think we need multiple arguments? I think the composite object instead of a state might be a really elegant solution. In combination with destructuring assignment it seems really powerful. Or am I missing something?

ellbee:
@speedskater  I went with multiple arguments because it makes `createSelector` backwards compatible with 0.0.2 and also because it matches the signature of `mapStateToProps`. 

I'm not sure I see the advantages of the composite object being that great, but I am open to be persuaded.

speedskater:
@ellbee I get your point. 

On the other hand using the composite object, the signature of a selector doesn't change at all. I think it should even work with the current version of reselect. 

I think having a single object as an argument encourages better composability and flexibility. 
E.g. instead of providing an object you could provide an array which will result in the same syntax you just proposed and is analogous to `mapStateToProps. `

```js
var mySelector = createSelector(
  ([ state, props ]) => state.value1,
  ([ state, props ]) => state.value2,
  ([ state, props ]) => props.value3,
  value1, value2, value3 => value1 + value2 + value3
);
mySelector([state, props]);
```
Using this pattern the arguments of `mapStateToProps` can directly be passed to such kind of selector using the spread operator.

In combination with the nice features you just added like different memoization functions i think it allows even more usage scenarios.

So for me the main advantages for this kind of pattern are:
* Keeping the selector simple
* Supporting many different usage scenarios not only tied to react or redux
* Keep it flexibel in case more than state and props should be taken into consideration

If we keep selector simple I think we should emphasize this flexiblity in the documentation. Hopefully this will help the users with `mapStateToProps` and in other advanced scenarios as well.

What do you think ?

ellbee:
@speedskater How does the array (or object) parameter work with the connect decorator? You couldn't just pass the selector straight in if you wanted to access the props right? I feel like I am missing something.

What if we don't take an opinion on this and have a variadic parameter list. Then it would be possible to use (state, props), ({state, props}), ([state, props]) or whatever else.

```js
export function createSelectorCreator(memoize, ...memoizeOptions) {
    return (...selectors) => {
        const memoizedResultFunc = memoize(selectors.pop(), ...memoizeOptions);
        const dependencies = Array.isArray(selectors[0]) ?
            selectors[0] : selectors;
        return (...args) => {
            const params = dependencies.map(dependency => dependency(...args));
            return memoizedResultFunc(...params);
        };
    };
}
```

speedskater:
@elbee You are right directly passing the selector without some kind of wrapper is impossible. (haven't thought about it this way).
I like your proposal of having a general purpose variadic parameter list, will make reselect in general more flexibel.

volrath:
Hi @ellbee and @speedskater,

I'm running some tests on the latest version of reselect yet, and the last commit (31a30bd) breaks the ability for `react-redux` to pass down `props` as the second argument to selectors.

You can find the reason here:
https://github.com/rackt/react-redux/blob/master/src/components/createConnect.js#L33-L43

As you can see, `react-redux` calculates whether to pass `props` or not based on the selector's `arguments`'s length, and having a Rest Parameter in the function definition makes its `length` be 0.

```
> (a => ({})).length
1
> ((a, b) => ({})).length
2
> ((...args) => ({})).length
0
```

I also took a look at all the other `react-redux`'s branches and it's the same situation.

I created a pull request that reverts the change: https://github.com/faassen/reselect/pull/40

Awesome work here btw, loving it :)
Hope this helps.

speedskater:
@elbee so we would have to explicitly create wrapper functions. So maybe keep the old single argument interface. At least it enforces the users to provide a wrapper function. imho this is not an ideal solution.
By the way I am offline now for the next 3 days. So the next response can take a bit longer.

ellbee:
@speedskater  I'm not sure I follow. I reverted the (state, props) so you can pass a selector directly to `connect`. You shouldn't need a wrapper function?

volrath:
I agree with @speedskater that it's not an ideal solution, I do like the flexibility of being able to pass down as many arguments as the calling interface wants, but in order to adhere to react-redux's `mapStateToProps` API we need to have at least two parameters in the returning function definition.

I would like to understand the reasoning/design decision behind react-redux's checking if `mapStateToProps` (as well as `mapDispatchToProps`) has more than 1 argument in order to pass down `props`, and not just passing it either way. Maybe @gaearon can shed a light on that.

With the current solution, we're forcing react-redux to give both `state` and `props` to the selectors (and their dependencies) even though `props` might not be needed. But then again, I don't understand why this would be a problem.

Another approach to have the flexibility of variadic parameter list while adhering to react-redux's API is to have both:

```javascript
export function createSelectorCreator(memoize, ...memoizeOptions) {
    return (...selectors) => {
        const memoizedResultFunc = memoize(selectors.pop(), ...memoizeOptions);
        const dependencies = Array.isArray(selectors[0]) ?
            selectors[0] : selectors;
        return (state, props, ...args) => {
            const params = dependencies.map(dependency => dependency(state, props, ...args));
            return memoizedResultFunc(...params);
        };
    };
}
```

Since:

```
> ((a, b, ...args) => {}).length
2
```

ellbee:
@volrath It just so happens that I was just playing around with exactly what you are proposing there on a local branch. I think this might be my preferred solution.

The single argument solution that I *think* @speedskater was advocating is my least preferred right now. I don't want to require the user to write a wrapper so it can be used with `connect`, which I think will be the primary use case for Reselect.

volrath:
@ellbee sounds good!

faassen:
I'm certainly not up to date on this issue, but I found it a bit odd that defaultMemoize has a valueEquals parameter. Shouldn't that be "createMemoize" or something? Though of course we should make sure people can create memoize functions without using it.

ellbee:
@faassen You mean rename `defaultMemoize` to `createMemoize`? I don't think `createMemoize` is quite right. `defaultMemoize` is itself the memioze function that takes the function to be memoized. The signature is similar to that of lodash.memoize. Maybe the API itself is a little off here?

ellbee:
In essence, `defaultMemoize` is our simple memoize function that has a non-configurable cache size of 1 and a configurable `valueEquals` function. So we need a name to sum that up. :-|

ellbee:
Oh, and the thinking behind giving `defaultMemoize` the valueEquals function was so that there was an easy upgrade path for `createSelectorCreator`.

#### Before

```js
import { isEqual } from 'lodash';
import { createSelectorCreator } from 'reselect';

const deepEqualsSelectorCreator = createSelectorCreator(isEqual);
```

#### After

```js
import { isEqual } from 'lodash';
import { createSelectorCreator, defaultMemoizeFunc } from 'reselect';

const deepEqualsSelectorCreator = createSelectorCreator(
  defaultMemoizeFunc,
  isEqual
);
```

speedskater:
Sorry for the late response but I have been on a short holiday for the last days.

@volrath, @ellbee As far as i undestand it is not that simple.
Enforcing to always have a selector of arity two has the disadvantage, that the selector will always be invoked on every property changes eventhough it does not depend on the properties at all. 

As far as i know, the reason that the signature check was implemented in react-redux was to prevent such calls. But i think this check should be done by the selector and not by the higher order component.

Another possiblity would be to create a selector with the same arity as the selector with the largest arity. The problem is that creating such a function is quite expensive and i would vote highly against it.

@gaeron what do you think. Could we remove the signature check in react-redux from the connect HOC and let the selector alone decide whether it  does resepect props or not? This would be equal to a reselect selector like this. 

```js
createSelector((state, props) => state, state => .....);
```

ellbee:
@speedskater Hey, no worries. Hope you had a good holiday.

I am aware that the selector is going to be called every time the props change whether we are using them or not. My thinking is that if the dependencies aren't using the props, a prop change won't cause the results function to be re-run because non of the parameters to the results function will have changed. For this reason I am presuming it won't be too big a performance hit. Am I missing something?

speedskater:
@ellbee you are right. I think it wouldn't be a performance hit at all. I even think, that the connect method should always provide the props parameter. As it is the duty of the selector to decide whether its result changes or not.

ellbee:
@speedskater (and everyone else), is there anything you think should prevent the release of 1.0.0? I'd like to get it done in the couple of days if everyone is happy.

speedskater:
@ellbee Thanks for your effort. From my point of view I think we could ship 1.0.0.

HeyImAlex:
Nice work!

volrath:
:+1: !

speedskater:
Great: :+1: !

PR #40, opened 2015-08-31
Reverts last commit.

volrath:
It's necessary to have at least two parameters in the returning function definition in order to play nice with `react-redux`'s `connect` function.


ellbee:
Whoops. Thanks @volrath!

PR #46, opened 2015-09-11
Added support for mapping selectors

PSpSynedra:
Adds the `composeSelectors` API-Function to create mapping selectors.

### DIFF
--- a/test/test_selector.js
+++ b/test/test_selector.js
@@ -1,5 +1,5 @@
 import chai from 'chai';
-import {createSelector, createSelectorCreator} from '../src/index';
+import {createSelector, createSelectorCreator, createMappingSelector} from '../src/index';
 
 let assert = chai.assert;
 
@@ -66,5 +66,28 @@ suite('selector', function() {
         assert.equal(called, 1);
         assert.equal(selector({a: 'A'}), 'A');
         assert.equal(called, 2);
+    });
+    test("default selector function", function() {
+        const selector = createSelector([state => state.a, state => state.b]);
+
+        let firstResult = selector({a: 1,b: 2});
+        assert.deepEqual(firstResult, [1,2]);
+        assert.strictEqual(selector({a: 1,b: 2}), firstResult);
+        assert.notStrictEqual(selector({a: 2,b: 2}),firstResult);
+        assert.notEqual(selector({a: 2,b: 2}), firstResult);
+        assert.deepEqual(selector({a: 2,b: 2}), [2,2]);
+    });
+    test("mapping selector", function() {
+        const selector = createMappingSelector({
+              x: state => state.a,
+              y: state => state.b
+        });
+
+        let firstResult = selector({a: 1,b: 2});
+        assert.deepEqual( firstResult, {x: 1, y: 2});
+        assert.strictEqual(selector({a: 1,b: 2}), firstResult);
+        assert.notStrictEqual(selector({a: 2,b: 2}),firstResult);
+        assert.notEqual(selector({a: 2,b: 2}), firstResult);
+        assert.deepEqual(selector({a: 2,b: 2}), {x: 2,y: 2});
     });
 });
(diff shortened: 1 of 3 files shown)


test/test_selector.js
```diff
@@ -189,4 +189,40 @@ suite('selector', () => {
...
+        assert.notEqual(selector({a: 2,b: 2}), firstResult);
+        assert.deepEqual(selector({a: 2,b: 2}), {x: 2,y: 2});
+    });
+    test("composing selectors with custom selector creator", function() {
+        let customEqualityFunctionCalls = 0;
+        const customSelectorCreator = createSelectorCreator(
+            defaultMemoize,
+            (a,b) => {
+                ++customEqualityFunctionCalls;
+                return JSON.stringify(a) == JSON.stringify(b)
```
ellbee: I can't figure out the reason behind counting the calls to the equality check function. Also, JSON.stringify(a) == JSON.stringify(b) can be replaced by a === b and the tests still pass.

test/test_selector.js
```diff
@@ -189,4 +189,40 @@ suite('selector', () => {
...
+              y: state => state.b
+        }, customSelectorCreator );
+
+        let firstResult = selector({a: 1,b: 2});
+        assert.deepEqual( firstResult, {x: 1, y: 2});
+        assert.strictEqual(selector({a: 1,b: 2}), firstResult);
+        assert.notStrictEqual(selector({a: 2,b: 2}),firstResult);
+        assert.notEqual(selector({a: 2,b: 2}), firstResult);
+        assert.deepEqual(selector({a: 2,b: 2}), {x: 2,y: 2});
+
```
ellbee: Do we need the notEqual and notStrictEqual checks on lines 222 and 223 if we have the deepEqual check on line 224?

ellbee:
Thanks @PSpSynedra, I have been trying this out and it is working great so far. I've just made a couple of line notes about the tests.

test/test_selector.js
PSpSynedra: That's right, these two checks are not necessary as you pointed out correctly. I will simply remove them.

test/test_selector.js
PSpSynedra: The stringify is simply to demonstrate that any custom equality function may be used.
The counting only ensures, that the equality function is not being called to often while checking the arguments for equality. But with the other tests still passing, this is no necessity either.
I am gonna simplify the case.

ellbee:
I wonder if `combineSelectors` is a better name than `composeSelectors` for this? It seems to me we are combining the selectors into a single selector rather than composing the selectors together.

PSpSynedra:
Yeah the decision was between the two of them and as we called it combineSelectors internally, well ;-)
But for me combineSelectors is also fine as it actually combines them into a map.
Although the map needs to be provided by the caller already so the actual combining needs to be done by the caller.

Hm *thinking*. 

Maybe createMappingSelector or createStructuredSelector would describe its functionality even better?

PSpSynedra:
No we called it composeSelectors of course, sorry.

ellbee:
Yes, I guess `combineSelectors` might not the right name either!

PSpSynedra:
What about createMappingSelector or createStructuredSelector as mentioned earlier?

ellbee:
Yeah, I agree that they more accurately describe what is going on so I think they are candidates. I have another possibility to throw out, how about `createKeyedSelector`?

PSpSynedra:
I'd rather prefer `createStructuredSelector`, as it allows you to create an arbitrarily structured selector like in:
```javascript
let selector = createStructuredSelector({
   subA: createStructuredSelector({
       selectorA: whatever
       selectorB: whatever
   }),
   subB: createStructuredSelector({
       selectorC: whatever
       selectorD: whatever
   })
});

let values = selector( state );
/* Expect values to be structured like:
   {
       subA: {
           selectorA: value
           selectorB: value
       },
       subB: {
           selectorC: value
           selectorD: value
       }
   }
*/
```

ellbee:
OK, let's go with that then. I'll merge this in later today.

ellbee:
@PSpSynedra, thanks a lot for working on this! It has been merged and released as v1.10.

I'm going to close the PR because I merged the commits locally so I could squash them.

PSpSynedra:
Cool, thanks

PR #49, opened 2015-10-01
Validate input selectors on selector creation

SpainTrain:
### Rationale

I ran into an issue while working on some selectors where an input selector was undefined.  However, the error is thrown only when the selector is used and does not provide a terribly useful stack trace.  Checking the input selectors on selector creation:

* Fails sooner, when the selector is originally created
* Provides us a stack trace directly to the problematic code

### Solution

A very basic `typeof` check to ensure than input selectors are functions.

### Changes

* Ensure input selectors are functions during selector creation
* Throw if invalid, allowing for a useful stack trace

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -37,6 +37,11 @@ export function createSelectorCreator(memoize, ...memoizeOptions) {
             return memoizedResultFunc(...params);
         };
 
+        if (!dependencies.every(dependency => typeof dependency === 'function')) {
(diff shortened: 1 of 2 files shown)


ellbee:
:+1: Thanks very much!

ellbee:
I'm going to write similar checks for createStructuredSelector and release both those checks and this PR in a new version, hopefully in the next couple of days.

PR #60, opened 2015-11-11
Improve createSelector title

weblancaster:
Although the duplicated title is intended it was confusing at first, moving to description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,9 +350,8 @@ When a selector is connected to a component with `connect`, the component props 
 ## API
 
 ### createSelector(...inputSelectors, resultFunc)
-### createSelector([inputSelectors], resultFunc)
 
-Takes one or more selectors whose values are computed and passed as arguments to `resultFn`.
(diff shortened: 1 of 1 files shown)


ellbee:
Yeah, good point. Thanks for the PR. I have no time at all to look at this for the next couple of days, but I intend to merge (I would just like to think if I want to tweak the wording a bit).

weblancaster:
Ok cool.. yeah this is no more than 2mins @ellbee 

thanks

mindjuice:
I would leave the extra function definition and just update the wording a bit to clarify:

    Takes one or more selectors (or an array of selectors) whose values are computed and passed as arguments to `resultFn`.

Alternatively to avoid the passive voice:

    Takes one or more selectors, or an array of selectors, computes their values and passes them as arguments to `resultFn`.

ellbee:
@weblancaster @mindjuice Thanks for the input, I've sort of ended up doing something in the middle.

PR #65, opened 2015-12-14
Typo in createSelectorCreator

compulim:


ellbee:
Hi @compulim, that's not a typo—we are creating a selector named `customSelector` by calling `customSelectorCreator` that we defined a few lines above. Thanks for taking the time to make a PR though, it's appreciated.

compulim:
@ellbee, Oh right, I overlooked at `customSelectorCreator` and think it was `createSelectorCreator` instead. :+1:

PR #68, opened 2015-12-21
Remove implication that state/props are inherent arguments.

bryanhelmig:
It feels very React-specific and unnecessary to label those arguments, so why not drop it?


ellbee:
Thanks for the PR but we do actually need to label the arguments. See here : https://github.com/rackt/reselect/pull/40/files

PR #72, opened 2016-01-14
Use loose mode for modules

SimenB:
Support IE8

Changing to `exports.__esModule = true` manually in the imported code makes it work. Without it, `Object.defineProperty` is used.

```js
Object.defineProperty(exports, '__esModule', {
  value: true
});
```
This is a no-op in core-js for IE8, so the `import`-transform in babel doesn't work.

And yes, I know IE8 isn't supported by MS anymore, but this is a small change.

Workaround for now is to use `const { createSelector } = require('reselect');`, but I don't want that :laughing:

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "url": "https://github.com/rackt/reselect/issues"
   },
   "scripts": {
-    "compile": "babel -d lib/ src/",
+    "compile": "babel --loose es6.modules -d lib/ src/",
     "lint": "eslint src test",
     "prepublish": "npm run compile",
     "test": "NODE_ENV=test mocha --compilers js:babel/register --recursive",


ellbee:
Cool, I don't see any downsides with supporting ie8 while its feasible. Thanks!

SimenB:
Great, thanks! We're merging a PR tomorrow using reselect. Would it be possible to publish this before that?

ellbee:
I'm on it!

ellbee:
v2.0.2 published to NPM

SimenB:
Awesome!

On Thu, 14 Jan 2016, 22:17 Lee Bannard <notifications@github.com> wrote:

> v2.0.2 published to NPM
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rackt/reselect/pull/72#issuecomment-171782411>.
>
-- 

--
Simen Bekkhus

PR #78, opened 2016-02-01
Fix issue with defaultMemoize and multiple arguments

WadePeterson:
Noticed an issue with `defaultMemoize` when you pass in an ordered subset of the memoized arguments, it will incorrectly returned the memoized value.

See test for an example (it fails without the argument length check).


ellbee:
Yep, I think you are right, good catch! Thanks for the fix, much appreciated. :+1:

PR #85, opened 2016-02-19
Check if "some" arg is not equal instead of checking if "every" arg is equal

dara76:
(changed the title to make it fit)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,7 +9,7 @@ export function defaultMemoize(func, equalityCheck = defaultEqualityCheck) {
     if (
       lastArgs !== null &&
       lastArgs.length === args.length &&
-      args.every((value, index) => equalityCheck(value, lastArgs[index]))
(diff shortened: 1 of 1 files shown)


ellbee:
Hey @dara76, thanks for the PR. I'm afraid I don't understand the reason for the proposed change.

dara76:
@ellbee - My thinking is it's a small performance improvement when `args` is different to `lastArgs` because if one argument fails the equality check then there's no need to check the rest of the arguments, we already know that we will need to call the function again.

ellbee:
`Array.every` returns false as soon as it finds an element that returns a falsy value from the predicate function, so I don't think this PR will result in a performance increase.

ellbee:
I think that `Array.every` reads better here so I am going to close this out, but thanks for taking the time to contribute, I do appreciate it.

dara76:
> Array.every returns false as soon as it finds an element that returns a falsy value from the predicate function

Of course it does, not sure what I was thinking.

PR #92, opened 2016-03-10
attach resultFunc to selector response to simplify isolated unit testing of composed selectors

HenrikJoreteg:
When working with a large redux store and composed selectors that ultimately depend on a tree of other selectors it would be nice to be able to have the option of writing a unit test for a selector that only tests what's *unique* to that selector (which is the result function). 

Simply attaching it to resulting function like this makes it easy to just call that function in our unit testing. 

Then we can easily test only the logic that's unique to that selector since we've already got tests for the other ones. Without this, we sometimes find ourselves having to re-create complex store state just so our result function can called with the arguments we're trying to test.

Exposing this on the result lets us do this without also exporting our result function from the module.

Thoughts? Would be happy to add whatever other docs/test for this, just wanted to float the idea first.


ellbee:
Genuine question: what do you see as the advantage of exposing the result function as a property instead of exporting the function?

HenrikJoreteg:
@ellbee fair question.

We have several selectors based on other selectors, but often the result function isn't useful on its own because it is dependent on other selectors.

We've got a file that exports a bunch of selectors.

If we could have access to the result function we could still write them like this:

```js
export const shouldFetchSelector = createSelector(
  firstSelector,
  secondSelector,
  (first, other) => first + second
)
```

Otherwise we'd have to export twice as many functions from the module, with no gain other than easier testing.

We'd have to do this: 

```js
export const addResult = (first, other) => first + second
export const shouldFetchSelector = createSelector(
  firstSelector,
  secondSelector,
  addResult
)
```

While this might be ok, I find it less readable and forces me to export a function that's otherwise useless for the app. 

Plus when you've got lots of selectors, it's nice to just export what your app actually cares about.

ellbee:
This kind of request keeps coming up. Personally, I haven't had any trouble with unit testing, I've just been passing data in and asserting on the results, but I accept that other peoples use cases might be more complicated than mine. Exposing the internals of the selector isn't really to my taste, but I don't know of a correctness/performance argument against adding properties to a function (and I have already done this with the `recomputations` property), so I think I am going to accept the PR. Does anyone reading this have a strong opinion that this is the *wrong* thing to be doing?

If we do this, I suggest we also add a property for the input funcs, which will make @tomatau happy I think!

tomatau:
This does make me happy. :+1: For large and complex state objects with composed selectors. It's a pain to have to couple selectors to state (in tests) that an implementation (another selector) has already abstracted! 

This is a good step in the right direction... although I'd still like to see a way to stub the return values of a subset of the other composed selectors that the selector under test is using.

```js
// source
const superComplexSelector = createSelector(
  [ firstComplex, secondComplex, simpleSelector ],
  (first, second, simple) => // stuff
)

// test, assuming firstComplex and secondComplex have already been vigorously tested.
firstComplex.stubReturn('some value for test condition')
secondComplex.stubReturn('some value for test condition')
// note, this test has no knowledge of simpleSelector
expect(superComplexSelector(state)).to.eql('happy')
```

ellbee:
Yeah, my thinking is that if you expose the actual input function array (dependencies in the code) on the selector you would be able to stub them.

tomatau:
That would definitely help! 

Although, it could be better? Still feels a bit coupled to implementation... you need to always either put the selectors to stub at the head or tail of that input array... or give the test knowledge of how many other input functions the selector depends on.

I'm not sure how to get around this problem as the references seem to get lost to the input functions once they go into createSelector.

ellbee:
@HenrikJoreteg There doesn't seem to be any objections to this so would you like to add some docs and tests? If you don't have time right now then that's cool, let me know and I'll just merge the PR and do the docs myself. Thanks for your patience!

markerikson:
Seems like there's a reasonable parallel with React Redux's generated wrapper component exposing `WrappedComponent`.

HenrikJoreteg:
@ellbee Cool, I should hopefully be able to add some docs/tests today.

HenrikJoreteg:
@ellbee ok, I think it's all set. Let me know if I missed anything or screwed anything up.

ellbee:
Looks great, thank you.

HenrikJoreteg:
@ellbee sweet, thanks! Planning to do a patch release by chance? :smile:

ellbee:
Sure, that would be useful I guess. :smile: Released!

PR #94, opened 2016-03-14
Reference typings in package.json

ianks:
This PR allows typescript users to simply `npm install reselect` to get typings support. For more information on how this works, check out this link: https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages

As a note, I gave @frankwallis the commit authorship, because he wrote the original commits for the typings. He was not involved in this PR, however. Sorry for any confusion!


ellbee:
Thank you!

PR #95, opened 2016-03-15
Pass all args to the selector function

mdebbar:
Why do we have to special-case `props`? Let's pass all the arguments to the dependencies **and** to the result function. I'm still new to Reselect so I'm curious to know if I'm doing something wrong here.

This change will allow us to create selectors that take arguments:
```js
const allBooks = (state) => state.books;

const oldBooks = createSelector(
  [allBooks],
  (books, yearLimit) => books.filter(book => book.publishingYear < yearLimit)
);

oldBooks(state, 1990); // recomputation (initial computation)
oldBooks(state, 1990); // from cache

const booksByAuthorId = createSelector(
  [allBooks],
  (books, authorId) => books.find(book => book.authorId === authorId)
);

booksByAuthorId(state, 123); // recomputation (initial computation)
booksByAuthorId(state, 123); // from cache
```

It can be used with `react-redux` as follows:
```js
const mapStateToProps = (state, ownProps) => ({
  oldBooks: oldBooks(state, ownProps.yearLimit),
  booksByAuthorId: booksByAuthorId(state, ownProps.authorId),
});

const MyConnectedComponent = connect(mapStateToProps)(MyComponent);
```

See [my comments](https://github.com/reactjs/reselect/pull/59#issuecomment-196584574) for more context.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -49,11 +49,11 @@ export function createSelectorCreator(memoize, ...memoizeOptions) {
       ...memoizeOptions
     )
 
-    const selector = (state, props, ...args) => {
+    const selector = (state, ...args) => {
       const params = dependencies.map(
-        dependency => dependency(state, props, ...args)
+        dependency => dependency(state, ...args)
       )
-      return memoizedResultFunc(...params)
+      return memoizedResultFunc(...params, ...args)
     }
 
     selector.recomputations = () => recomputations


ellbee:
Props *looks* special cased because it won't work with React Redux otherwise. See #40 and #68. I'll add a comment to the code about this because it is highly non-obvious.

The reasoning as to why we don't currently support selectors that take arguments can be found [here](https://github.com/reactjs/reselect#q-how-do-i-create-a-selector-that-takes-an-argument).

mdebbar:
If I'm understanding the suggestions correctly, there is a big issue that they introduce:
```js
const allBooks = (state) => state.books;
const oldBooks = createSelector(
  allBooks,
  (books) => memoize(
    yearLimit => books.filter(book => book.publishingYear < yearLimit)
  )
);

const getOldBooks = oldBooks(state);

getOldBooks(1990); // works fine.

// State gets updated...

getOldBooks(1990); // returns out-dated results :(
```
The problem is that memoized function that I created is recording an old state and never getting the new state. Unaware users will definitely fall for this and have nasty bugs.

I've been reading the issues and discussions about dynamic arguments but I still don't understand the **_reason_** for not supporting them. I understand that there are workarounds for some cases (e.g. putting the dynamic argument in the state). But that doesn't solve all the use cases.

I guess what I'm looking for is: what's exactly wrong with this PR? Does it introduce any issues that I'm not aware of? Does it add a lot of concepts that make the library difficult to learn/use?

If you read through [my comments](https://github.com/reactjs/reselect/pull/59#issuecomment-196577520) in the [closed issue](https://github.com/reactjs/reselect/pull/59), I've explained why all the workarounds don't work in my case.

PR #96, opened 2016-03-17
Fix FAQ link

joshkel:


ellbee:
Ok, this is better. Thanks.

PR #106, opened 2016-04-04
Add bindSelectorCreators

### LINKED ISSUE #102: bindSelectorCreators to state
Hi. It would be useful to have `bindSelectors` function which can bind `getState` like
Something like https://github.com/reactjs/redux/blob/master/src/bindActionCreators.js
What do you think?

TrySound:
Fix #102


ellbee:
Closing as #102 is closed.

PR #107, opened 2016-04-05
Clarify when it's necessary to copy selectors

courthead:
Previously, it was unclear for people skimming the docs that the advice found in the `Sharing Selectors Across Multiple Components` section only applies for those trying to solve the specific problem mentioned in the previous section. As a result, I'm sure some people were padding private copies of selectors to different components even in situations where they didn't need to. This makes the documentation more clear.


ellbee:
Thank you!

PR #112, opened 2016-04-16
Add UMD build command

### LINKED ISSUE #99: Add browser compatible version
1 - when I ```npm install reselect``` npm gave me a peer dependency error for react-redux. I dont use react for my project and thankfully reselect library works without the react dependency. Please remove this dependency error from the npm dependency declarations.

2 - Please provide a bundled non ES6 version of the library, so people can also download them without ever needing an ES6 transpiler. the whole code base isn't that big. Ideally every function could be namespaced to Reselect global var. Example: Reselect.createSelector();

thank you for creating an awesome library, please make sure it's accessible to non-react-hipsters too ;)

The example project that uses reselect without react is here: https://github.com/izelnakri/riotjs-experiment

npbee:
Adds a command in `package.json` to build a browser-ready UMD file to a `dist`
folder.  All files are compiled to a `reselect.js` file and attached to a global
variable name of Reselect.

Closes #99

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
 lib
 coverage
+dist
 
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   "files": [
     "lib",
     "src",
+    "dist",
     "README.md",
     "LICENSE",
     "package.json"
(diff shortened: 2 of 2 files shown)


npbee:
Gah, sorry.  I have a pesky `.DS_Store` file in here.  Going to fix and force push up.

ellbee:
Thanks very much, appreciate it!

ellbee:
Published to npm as v2.4.0

PR #115, opened 2016-04-19
Typed the `props` argument on the selectors created by `createSelector`, and added typings test cases.

geon:
The test is done by simply running `tsc` in the `/tests/typings/` folder. If it compiles, everything is fine.


geon:
The tests should probably be runnable by Travis CI. Anyone know how to fix?

src/reselect.d.ts
```diff
@@ -5,22 +5,22 @@
 
 declare module Reselect {
 
-    type Selector<TInput, TOutput> = (state: TInput, props?: any) => TOutput;
+    type Selector<TInput, TProps, TOutput> = (state: TInput, TProps, props?: TProps) => TOutput;
```
HeyImAlex: Was this a typo? That looks like you accidentally included a required implicit any parameter named TProps.

geon: Oops. Yes.

test/typings/common.ts
```diff
@@ -0,0 +1,22 @@
+
+export type RootState = {
+	items: {[key: string]: {
```
threehams: Can you convert tabs to spaces to match the other files? It's a small thing, but it'll help avoid whitespace changes on future diffs.

test/typings/should_not_compile/index.ts
```diff
@@ -0,0 +1,26 @@
+import {createSelector} from '../../../src/reselect.d.ts';
+import * as common from '../common';
+
+
+const implicitlyTypedSelector = createSelector(
```
threehams: Do you have any examples of negative tests written for type definitions? I think they could be very useful, but checking for _any_ compiler error in any test won't scale very well.

If there's no well-known solution, I'd rather leave these out to simplify integration with CI.

test/typings/should_not_compile/tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "allowUnreachableCode": true
```
threehams: This only works in TypeScript 1.8, so we won't be able to test backwards compatibility. Still, having sanity-check tests are definitely better than nothing, so it's worth merging in - I can open a ticket for definition test coverage.

src/reselect.d.ts
```diff
@@ -5,22 +5,422 @@
...
-    function createSelector<TInput, TOutput, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14>(selector1: Selector<TInput, T1>, selector2: Selector<TInput, T2>, selector3: Selector<TInput, T3>, selector4: Selector<TInput, T4>, selector5: Selector<TInput, T5>, selector6: Selector<TInput, T6>, selector7: Selector<TInput, T7>, selector8: Selector<TInput, T8>, selector9: Selector<TInput, T9>, selector10: Selector<TInput, T10>, selector11: Selector<TInput, T11>, selector12: Selector<TInput, T12>, selector13: Selector<TInput, T13>, selector14: Selector<TInput, T14>, combiner: (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, arg7: T7, arg8: T8, arg9: T9, arg10: T10, arg11: T11, arg12: T12, arg13: T13, arg14: T14) => TOutput): Selector<TInput, TOutput>;
+    type Selector<TInput, TProps, TOutput> = (state: TInput, props?: TProps) => TOutput;
```
threehams: Thank you for putting whitespace changes in a separate commit!

geon: Anytime. I'm not a savage.

test/typings/should_not_compile/index.ts
geon: I searched for this, but could find nothing. I just wrote a SO question about it. Maybe something comes from that: https://stackoverflow.com/questions/37927730/how-do-i-write-negative-tests-for-typescript-typings

geon: If you can't get them to work with the CI, just leave them out. But keep the tests so they can be run manually. Better than nothing, I think.

test/typings/common.ts
geon: Done: https://github.com/geon/reselect/commit/8a9b4e8983f531e15040e188df504a68d2da12a9

test/typings/should_not_compile/index.ts
threehams: OK. I've added an issue to get definition tests running on CI, which I'll take. If you get any responses on negative testing strategies, feel free to send another pull request.

ellbee:
🎉 Thanks for this @geon!

And thanks @threehams for reviewing and merging. Much appreciated!

PR #116, opened 2016-04-20
Add jsnext build

### LINKED ISSUE #103: jsnext:main
This field should point source code only with untranspilled modules. In current case for example rollup users should process all deps with babel which is too long or prebundle it separatly.
https://github.com/reactjs/reselect/blob/master/package.json#L6

You can see example here
https://github.com/reactjs/redux/blob/master/package.json#L6

npbee:
Adds a build step for the `jsnext:main` which compiles all features _except_ for modules to an `es` folder.  The previous `compile` command is now the `commonjs` build which ends up in the `lib` directory.

Fixes #103

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,12 +4,16 @@
     "transform-es2015-arrow-functions",
     "transform-es2015-block-scoping",
     "transform-es2015-function-name",
-    ["transform-es2015-modules-commonjs", { "loose": true }],
     "transform-es2015-parameters",
     "transform-es2015-spread",
     "transform-es2015-template-literals"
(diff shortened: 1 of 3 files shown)


npbee:
@ellbee, tried my hand at the `jsnext:main` build.  Let me know if any troubles!

TrySound:
@npbee Looks good. Thanks a lot.

ellbee:
Thanks @npbee! If @TrySound says its good, then I'm happy to merge. I'm away from my computer for the rest of the day, but I'll publish to npm tomorrow.

PR #119, opened 2016-04-23
Fix exampleState in readme

jacobrask:
The state passed to `shopItemsSelector` did not have a `shop` property.


ellbee:
Yep, you're right! Thanks

PR #123, opened 2016-04-28
fix TS module resolution

bbenezech:
I couldn't use npm Reselect typings as I do with eg Immutable.js
TS would not find the module.
It turns out you need to declare a module whose name match exactly the import's name, "" included.
I tested it, it works®

A good reference: https://github.com/facebook/immutable-js/blob/master/dist/immutable.d.ts

### DIFF
--- a/src/reselect.d.ts
+++ b/src/reselect.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: Frank Wallis <https://github.com/frankwallis>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
 
-module Reselect {
+declare module Reselect {
 
     type Selector<TInput, TOutput> = (state: TInput, props?: any) => TOutput;
 
--- a/src/reselect.d.ts
+++ b/src/reselect.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: Frank Wallis <https://github.com/frankwallis>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
 
-declare module Reselect {
+module Reselect {
 
     type Selector<TInput, TOutput> = (state: TInput, props?: any) => TOutput;
 
@@ -31,4 +31,6 @@ declare module Reselect {
     function createSelectorCreator(memoize: Memoizer, ...memoizeOptions: any[]): any;
 }
 
-export = Reselect;
+declare module "reselect" {
+  export = Reselect;
+}


threehams:
Using TypeScript 1.8, with the following import syntax and .tsconfig, I can compile without problems on master, but see errors when using the changes in your branch:

```
```
or 
```
import reselect = require('reselect');
```

.tsconfig
```
{
  "compilerOptions": {
    "allowJs": true,
    "experimentalDecorators": true,
    "jsx": "preserve",
    "module": "commonjs",
    "outDir": "tsDist",
    "target": "es2015",
    "moduleResolution": "node"
  },
  "exclude": [
    "node_modules",
    "server.js",
    "serverConfig.js",
    "webpack.config.js"
  ]
}
```

With this pull request, I see an error:
```
error TS2656: Exported external package typings file 'node_modules/reselect/src/reselect.d.ts' is not a module. Please contact the package author to update the package definition.
```

Can you give information on your environment - TS version, .tsconfig settings, and import / require syntax?

bbenezech:
@threehams Does it work with immutable ?

I just basically used their project as a template as they should know better.

threehams:
Again, can you post reproduction steps and version/environment/code? The ability to read definition files from node_modules was added in Typescript 1.6, so version matters a lot here, as well as es5/6 target, module setting, whether you're using Babel in addition to Typescript, etc.

Immutable creates a module, exports everything from that, then re-exports it using the final string name. This makes it backwards-compatible with 1.5 and earlier when used outside node_modules, but I don't think we're not worried about doing that here. If we're targeting 1.6+, it's simpler to use top-level exports and avoid declaring modules at all:

```
// Type definitions for reselect v2.0.2

export type Selector<TInput, TOutput> = (state: TInput, props?: any) => TOutput;

export function createSelector<TInput, TOutput, T1>(selector1: Selector<TInput, T1>, combiner: (arg1: T1) => TOutput): Selector<TInput, TOutput>;
export function createSelector<TInput, TOutput, T1, T2>(selector1: Selector<TInput, T1>, selector2: Selector<TInput, T2>, combiner: (arg1: T1, arg2: T2) => TOutput): Selector<TInput, TOutput>;
...etc.
```

Information on writing definition files can be really tricky to find, and a lot of it is buried in comments in issues and pull requests.

https://github.com/Microsoft/TypeScript/issues/247#issuecomment-141133516

ellbee:
As we can't reproduce this, I am going to close due to inactivity.

Issue #134, opened 2016-06-16
Build scripts fail in Windows environment.


threehams:
There are some OS-specific command in the npm scripts which fail in Windows. Luckily, there are better-npm-run, cpy, and mkdirp to take care of it.

I can fix this in a pull request once I verify that the output is identical across Windows + OS X.

PR #135, opened 2016-06-16
Change npm scripts to work cross-platform.

threehams:
Fixes issue #134, adds three small dev-only dependencies. I've diffed compiled files on Windows 10 and OS X 10.10.5.

Whitespace changes on authors in package.json were done by npm automatically (Node 6.2.1).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ lib
 coverage
 dist
 es
-
+.vscode
+.idea
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
       }
     },
     "compile:commonjs": {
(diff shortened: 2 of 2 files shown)


threehams:
Hm. cpy requires Node 4+. I can look for an alternative if you'd like compilation to work in 0.12 and earlier.

ellbee:
As 0.12 is an LTS version it would be good to support it if possible. Can we switch out `cpy` with `ncp`?

threehams:
Done. Build now passes. Want me to squash commits?

ellbee:
Looks good, thanks!

PR #144, opened 2016-07-02
Cancel args memorization if func throws exception

zandroid:
The issue is when `func` throws exception the last `args` already saved and next calling of memoized func will return `null` instead of recalculating `func` again. If we save `args` after calling of `func` we save `args` only for "good" result and don't for exceptions.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,8 +13,8 @@ export function defaultMemoize(func, equalityCheck = defaultEqualityCheck) {
     ) {
       return lastResult
     }
-    lastArgs = args
     lastResult = func(...args)
+    lastArgs = args
     return lastResult
   }
 }
--- a/test/test_selector.js
+++ b/test/test_selector.js
@@ -79,6 +79,36 @@ suite('selector', () => {
     )
     assert.equal(selector({ a: 1, b: 2 }, { c: 100 }), 103)
   })
+  test('recomputes result after exception', () => {
+    let called = 0;
+    const selector = createSelector(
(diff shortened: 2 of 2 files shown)


ellbee:
Hey @zandroid, would it be possible to add a test case that fails on the current version and passes with your fix please?

zandroid:
@ellbee, sure, have added two tests.

zandroid:
> So either it never calculated a result without throwing (returning the default value of null)

It's exactly the issue that I have in my code. But anyway I think selector should not memoize (default or previous) result for current args when `func` throws exception.

There may be other issue, not for default `null` value, for example:
```javascript
const selector = createSelector(state => state.a, (a) {
  if (a > 1) throw new Error()
  return a;
})
selector({ a: 1 }) // => 1 - OK
try {
  selector({ a: 2 }) // => throw error - As expected
} catch (e) {}
selector({ a: 2 }) // => 1 - What?!
```

ellbee:
Ok, thanks! So the problem is that it doesn't throw the second time—the return value being `null` is not relevant. It seems strange to have your selectors throwing errors, but I think I agree that the behaviour in this PR makes more sense.

ellbee:
@zandroid Released in v2.5.2. Thanks for the PR!

zandroid:
A lot of thanks. I agree that throwing exceptions from selector is not usual case, but we use it in our selectors to detect incorrect id of requested entity and show 404, 403 or 401 pages.

PR #158, opened 2016-08-08
Pass argument's index to equalityCheck function

bsideup:
will allow to conditionally apply different checks on different arguments, i.e. defaultEqualityCheck for everything, but `_.isEqual` in case of non-immutable argument


ellbee:
Hey @bsideup, thanks for the PR! I don't see any reason why we shouldn't merge this, but I'm going to leave it open for the moment and aim to make it part of the v3 release.

ddemolina:
Hi, I can't see this PR merged in master branch. Is there any problem?
Thank you

PR #161, opened 2016-08-18
Updated lodash memoized example function with initialValue

dpwrussell:
I think it is probably sensible to have an empty string as the `initialValue` of this example reducer, otherwise the first argument in the array is not stringified.

I know it's just an example, but as I'd copied and pasted this as a starting point for something a little more complicated I ran into some problems (which I traced back to this) and I'm guessing other people might do the same.

Cheers


ellbee:
Thank you!

PR #167, opened 2016-09-08
Added the 'resultFunc' property to the type definition

lukevenn:
I was getting type errors when using the 'resultFunc' property in tests as it was not declared on the Selector type. As a result, I've changed the Selector type to be an interface to allow for extension. From this I've created an OutputSelector interface that is extends it to add an optional 'resultFunc' property. Having it as an optional property prevents any errors where it is not being used.

To allow type checking to occur on the 'resultFunc' it is passed a generic TCombiner value that will be any one of the Combiner interfaces (Combiner1 through Combiner14). The reason for creation of the Combiner interfaces was to allow it's re-use in both the 'createSelector' definitions as well as the OutputSelector. Obviously the reason for having 14 Combiner interfaces is purely to allow it to be used across the 14 'createSelector' definitions. I couldn't think of a better way to do this and maintain type checking so I'm open to any ideas for improvement.

I also changed the module definition to be a namespace as I was seeing the following deprecation warning about using the module syntax - "The internal 'module' syntax is deprecated, use the 'namespace' keyword instead." It seemed like a good idea to update this at the same time but if you'd rather leave it as a module then please ignore.

### DIFF
--- a/src/reselect.d.ts
+++ b/src/reselect.d.ts
@@ -1,26 +1,89 @@
 // Type definitions for reselect v2.0.2
-// Project: https://github.com/reactjs/reselect
+// Project: https://github.com/rackt/reselect
 // Definitions by: Frank Wallis <https://github.com/frankwallis>
 // Definitions: https://github.com/borisyankov/DefinitelyTyped
 
-declare module Reselect {
+declare namespace Reselect {
 
-    type Selector<TInput, TOutput> = (state: TInput, props?: any) => TOutput;
+    interface Combiner1<T1, TOutput> {
+      (arg1: T1): TOutput;
+    }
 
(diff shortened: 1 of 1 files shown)


ellbee:
Hey @lukevenn, thanks for the PR. We have a more up to date version of the typings in the 3.0.0 branch, can you check that your changes are still valid against that branch? @threehams Is this something we should add for 3.0.0?

lukevenn:
Thanks @ellbee. From looking at 3.0.0 the changes will still be valid after adding in the new TProps generic.

tamlyn:
I've tested these new typings and they work. What is required to get this merged?

alex3165:
@lukevenn Let's open a new PR including this changes against the branch 3.0.0. I will close this PR, we will open a new one soon.

PR #172, opened 2016-09-16
Make defaultMemoizer store the latest args if the equality test passes

yoni-tock:
Closes https://github.com/reactjs/reselect/issues/170

_For posterity_
Previously, not updating `lastArgs` when the equality test passed
meant that if an object goes from shallow equal -> deep equal ->
shallow equal accross updates, that transition back to shallow equal
would continue to require deep equality checks. This could lead
to large slowdowns for complex objects. See the github issue
for details.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,13 +7,12 @@ export function defaultMemoize(func, equalityCheck = defaultEqualityCheck) {
   let lastResult = null
   return (...args) => {
     if (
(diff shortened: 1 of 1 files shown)


ellbee:
Great, thank you!

yoni-tock:
Thanks for accepting the PR! 🎉

PR #173, opened 2016-09-18
Improve defaultMemoize performance

pesho:
`defaultMemoize` creates an unnecessary inline function on (nearly) every invocation of the returned function. This affects both speed and memory usage. The function has been moved to the outer scope.

A very simple benchmark shows ~ 30% perf improvement in Chrome and ~ 150% in Firefox: https://www.measurethat.net/Benchmarks/Show/259/9/reselect-defaultmemoize


ellbee:
Hey, thanks for the PR. I am dubious that this will give any noticeable speed improvement outside of micro-benchmarks, but it can't hurt.

PR #190, opened 2016-11-10
Rename makeGetVisibleTodos example in README.md

spen:
Hi All! 👋

First off, thanks for the awesome work! 🙇

Some of us have been having a [discussion](https://github.com/mxstbr/react-boilerplate/pull/1205) over at React Boilerplate ([stemming from here](https://github.com/mxstbr/react-boilerplate/issues/882)).

>I'm not particularly sold on the naming: `makeSelectLocationState`.
I feel this is a little confusing since `make` & `select` are both verbs.

As you can see in the PR, I've pointed out that this function name starts with 2 verbs, which seems a little weird to me.

I thought I'd open this PR as a place to see what others (especially the authors) think about this naming and wether my suggestion, or an alternative, might make a little more sense :)

TIA ❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ A selector created with `createSelector` only returns the cached value when its 
 
 To share a selector across multiple `VisibleTodoList` components while passing in `props` **and** retaining memoization, each instance of the component needs its own private copy of the selector.
 
-Let’s create a function named `makeGetVisibleTodos` that returns a new copy of the `getVisibleTodos` selector each time it is called:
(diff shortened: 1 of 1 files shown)


jimbolla:
It's awkward, but it's a function that returns a function. And it's a somewhat established convention. And the alternative isn't amazing and perfect.  I think it'd be better to just leave it as is.

spen:
@jimbolla Fair points! :) 
Happy for this to be closed if there's no feedback from others 👍

Issue #146, opened 2016-07-04
Debugging errors in selectors


kjeske:
Today I came across strange problem with debugging my redux app. I started getting exceptions like:
```
TypeError: Cannot read property 'replaceChild' of null
```
I spent some time on investigation where is the issue, as the debugger didn't say anything about the real cause (replaceChild error was of course just a result of another thing). After some time I noticed that my selector contained a bug:

```
export const getSelectedItem = createSelector(
  [getItems, getSelectedItemId],
  (items, itemId) => {
    return items.finds(p => p.id === itemId)
                     ^
                    bug
  }
)
```

I used `finds` instead of `find`. Wow, that was hard to debug and my app is really small right now. I started wondering what will happen in bigger app. I know that in many cases the tests will catch such problems, but sometimes when we are playing with the code it's easy to stuck in a mud for long hours.

Is that just my case or it is common to have such debugging problems?

ellbee:
I'm going to close this as I don't think there is anything actionable for Reselect here, as this would happen in a plain selector also. I think bugs manifesting themselves far from their source is a weakness of dynamic languages. You might want to try unit testing your selectors as you go, or consider using Flow (or Typescript):

```js
/* @flow */
import { createSelector } from 'reselect';

const getItems = state => state.items
const getSelectedItemId = state => state.selectedItemId

const getSelectedItem = createSelector(
  [getItems, getSelectedItemId],
  (items : Array<any>, itemId) => {
    return items.finds(p => p.id === itemId)
  }
)
```
```sh
% flow
src/index.js:10
 10:     return items.finds(p => p.id === itemId)
                      ^^^^^ property `finds`. Property not found in
 10:     return items.finds(p => p.id === itemId)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Array


Found 1 error
```

kjeske:
Thanks for the answer!

josepot:
@kjeske Please have a look at the #191, do you think that something like that would help?

PR #191, opened 2016-11-11
[Enhancement] Add 'subscribeToErrors' for ease of debugging

josepot:
I think that a common problem when working with reselect is that often when the result function of a selector throws an error, that error manifests itself too far from its source and the stack-trace is not very useful. For example, consider getting the following error:

> Cannot read property 'id' of undefined

In a large project where there are many different selectors that are trying to access a property with that name... It would be very difficult to find what happened. If we could just know: which is the function that crashed and the parameters that it received, that alone would be very useful.

I agree with the following statement:

> bugs manifesting themselves far from their source is just a weakness of dynamic languages

that I read in #146 , but at the same time I think that we could do something to ease debugging with reselect. That's why I think that the new API function that I'm proposing in this PR could be pretty helpful.

Its typical use would be in the DEV environment. For instance, adding something like this in the entry file of the project:

```
import { subscribeToErrors } from 'reselect';

if (DEV) {
  subscribeToErrors((error, resultFn, args, dependencies) => {
    console.log('A selector just crashed.');
    console.log('The error message is', error.message);
    console.log('The code of the result function that crashed is', resultFn.toString());
    console.log('The arguments that it received were', args);
    console.log('The code of its dependencies is:', dependencies.map(fn => fn.toString()).join('\n\n'));
  });
}

...
```

Would be a huge help in order to debug issues in the result functions of the selectors.

BTW: There were some errors with the typescript tests: meaning that with the [current version](https://github.com/reactjs/reselect/tree/fb16b0289e3ab9db0316907d72f3d366bf47f634), running `npm run test:typescript` crashes. Basically what happens is that it won't compile the ts file that's supposed to compile successfully. I've fixed that in order to make sure that the typescript tests for the new function are passing. Please let me know if that's ok.


src/index.js
```diff
@@ -44,7 +46,12 @@ export function createSelectorCreator(memoize, ...memoizeOptions) {
     const memoizedResultFunc = memoize(
       (...args) => {
         recomputations++
-        return resultFunc(...args)
+        try {
+          return resultFunc(...args)
+        } catch (e) {
+          errorSubscriptions.emitError(e, resultFunc, args, dependencies)
```
JamesHageman: Would there be any way to attach the list of dependencies directly to the error object? It would be nice to get these details without having to subscribe externally.

josepot: That's a good point.

I don't like the idea of hijacking the Error. However, you are right, it could be possible to create a higher order function like this one:

```

const catchableSelector = (...args) => {
  const selector = createSelector(args);
  return (...state) => {
    try {
      return selector(...state);
    } catch (error) {
      const resultFunc = args.slice(-1);
      let dependencies = args.(0, -1);
      dependencies = dependencies.length === 1 && Array.isArray(dependencies[0]) ?
        ...dependencies[0] :
        dependencies;
      // DO WHATEVER
    }
  }
};

export default catchableSelector;
```

And then in your project you should always use the `catchableSelector` instead of using directly the `createSelector`... It's an option. I am not a big fan of this approach, but it's an option. One inconvenient with this approach is that you wouldn't be able to have access to the parameters that were received by the result function of the selector that crashed. Which is one of the most valuable pieces of information that you would get if this PR got accepted.

renanbandeira:
@josepot  why did you closed this PR? It looks an excellent approach and I just used your fork to handle my errors.

PR #193, opened 2016-11-21
Remove excess asigment of lastArgs

Shastel:
Assign ```lastArgs``` only if ```equalityCheck``` returns ```false```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -13,8 +13,8 @@ export function defaultMemoize(func, equalityCheck = defaultEqualityCheck) {
       !args.every(isEqualToLastArg)
     ) {
       lastResult = func(...args)
+      lastArgs = args
     }
-    lastArgs = args
     return lastResult
   }
 }


jimbolla:
I think it's better the way it was. Oftentimes the equality check can be more performant this way. For example if the equality check is a shallow equals, usually that's implemented by first doing a strict equals check followed by iterating over the properties. If sequential calls to your memoizer pass the same strictly equal object, that can alleviate the need to do the expensive shallow equals checking.

johnhaley81:
@jimbolla the arguments are spread into the function when it's called so it will never be `===`.

jimbolla:
@johnhaley81 That's true for the array itself, but still applies to the items in the array.

johnhaley81:
@jimbolla yes but the above change doesn't affect that. It only affects the referential equality check on the array itself and does not affect the items in the array.

jimbolla:
That's not correct. If you don't assign lastArgs even when equality passes, then you're not capturing its contents into any variable for the next run. This is relevant if you pass shallow equals as your equalityCheck, for example. Take the following example:

```js
function shallowEquals(a, b) {
   if (a === b) return true;

   // do relatively-expensive prop comparisons here.
}

const memo = defaultMemoize(someFunc, shallowEquals);

const p1 = { foo: 1, bar: 'react' };
const p1Clone = { ...p1 };

const p2 = { fizz: 2, buzz: 'redux' };
const p2Clone = { ...p2 };

const result1 = memo(p1, p2);
const result2 = memo(p1Clone, p2Clone);
const result3 = memo(p1Clone, p2Clone);
```

With the change proposed in this PR, the computation for `result3` would be more expensive because `shallowEquals` would have to work harder because the values in lastArgs would be [p1, p2], and not `[p1Clone, p2Clone]`.

Shastel:
Hi @jimbolla. As far as i understand 
```js
const p1 = { foo: 1, bar: 'react' };
const p1Clone = { ...p1 };
p1  ===  p1Clone // -> false
```
Could you please give more relevant example for me

jimbolla:
@Shastel That's true if you use `defaultEqualityCheck` but not if you pass a custom value for `equalityCheck` such as a `shallowEquals` function.

johnhaley81:
@jimbolla but `lastArgs` is never passed into `defaultEqualityCheck`

jimbolla:
It's values are. [On line 8](https://github.com/Shastel/reselect/blob/9f506b11c2108440b683711b89c33fa060a6e7ab/src/index.js#L8)

jimbolla:
I took my example code above and turned in into a full test in PR #195. You'll see that it fails if you make this change.

Shastel:
@jimbolla thanks, i'm get your point

PR #194, opened 2016-11-23
Speedup cache hit performance of selectors

johnhaley81:
ES6, object/array spreading and the like make the code much more readable and easier 
to reason about.

However, this trade-off comes at the cost of performance. When running a selector, a copy of the
parameters were cloned many times which added a trivial amount of overhead for a single call. 
However if a selector was called thousands of times (e.x. a large list of items) then the trivial
amount of overhead added upwards of around 1 second to all selector calls regardless of cache
hits/misses.

Normally, readable code > super optimized code but selectors can be called very often and the
overhead associated with those calls has to be minimal.

The first commit in this PR adds a test that calls a basic selector a million times. Before the
performance increase it takes ~2.5 seconds on my machine. Afterwards it takes ~140 milliseconds.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,19 +2,35 @@ function defaultEqualityCheck(a, b) {
   return a === b
 }
 
+function areArgumentsShallowlyEqual(equalityCheck, prev, next) {
+  if (prev === next) {
+    return true
+  }
+
+  if (prev === null || next === null || prev.length !== next.length) {
+    return false
+  }
+
+  const length = prev.length
+  for (let i = 0; i < length; i++) {
+    if (!equalityCheck(prev[i], next[i])) {
+      return false
+    }
+  }
+
+  return true
+}
+
 export function defaultMemoize(func, equalityCheck = defaultEqualityCheck) {
   let lastArgs = null
   let lastResult = null
-  return (...args) => {
-    if (
-      lastArgs !== null &&
-      lastArgs.length === args.length &&
-      args.every((value, index) => equalityCheck(value, lastArgs[index], index))
-    ) {
-      return lastResult
+
+  return function () {
+    if (!areArgumentsShallowlyEqual(equalityCheck, lastArgs, arguments)) {
+      lastResult = func.apply(null, arguments)
     }
-    lastResult = func(...args)
-    lastArgs = args
+
+    lastArgs = arguments
     return lastResult
   }
 }
@@ -27,7 +43,7 @@ function getDependencies(funcs) {
       dep => typeof dep
     ).join(', ')
     throw new Error(
-      `Selector creators expect all input-selectors to be functions, ` +
+      'Selector creators expect all input-selectors to be functions, ' +
       `instead received the following types: [${dependencyTypes}]`
     )
   }
@@ -42,18 +58,22 @@ export function createSelectorCreator(memoize, ...memoizeOptions) {
     const dependencies = getDependencies(funcs)
 
     const memoizedResultFunc = memoize(
-      (...args) => {
+      function () {
         recomputations++
-        return resultFunc(...args)
+        return resultFunc.apply(null, arguments)
       },
       ...memoizeOptions
     )
 
-    const selector = (state, props, ...args) => {
-      const params = dependencies.map(
-        dependency => dependency(state, props, ...args)
-      )
-      return memoizedResultFunc(...params)
+    function selector() {
+      const params = []
+      const length = dependencies.length
+
+      for (let i = 0; i < length; i++) {
+        params.push(dependencies[i].apply(null, arguments))
+      }
+
+      return memoizedResultFunc.apply(null, params)
     }
 
     selector.resultFunc = resultFunc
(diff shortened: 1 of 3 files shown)


jimbolla:
I haven't fully reviewed the code, but I support this in concept, and at quick glance it looks fine. I agree that performance is important for Reselect, because of how it's used. Also, used named functions over arrow functions is very useful for perf testing because seeing 'anonymous function' in Chrome dev tools makes it harder to see what's going on.

johnhaley81:
@jimbolla I added a top level memoizer on all selectors to stop recursively checking dependencies if referentially equal arguments are passed into a selector in commit https://github.com/reactjs/reselect/pull/194/commits/6d13eb47a33e6cc7d46793db5f482acb675140b6. Any thoughts?

FYI this brings the cache perf down 2 orders of magnitude now (~60ms on the test case).

johnhaley81:
@ellbee @jimbolla is there anything else you guys would like in this PR? It's been a few months and no word.

jimbolla:
It looks fine to me, but this isn't my project.

johnhaley81:
Ahh, my apologies @jimbolla.

/cc  @mjw56 @threehams @zandroid @1ven (copied from `AUTHORS`)

ellbee:
Sorry @johnhaley81, been busy on other things, but I appreciate your efforts here. I have no real objections to merging this, but I wonder how much of an impact these changes will make outside of micro-benchmarks? I guess it can't hurt though. Does anyone else have an opinion on this? I am thinking that I'll make time to release v3 this week now that the TypeScript typings have been merged, so will merge this before the v3 release if there is a demand.

johnhaley81:
No worries, and sorry about directly pinging everybody. Admittedly the test case is super simple and a micro-benchmark but it is a fundamental case for any and all selectors created with reselect. The numbers only get worse as you increase both the breadth and depth of the selector call tree. Especially if you have a lower level selector that is called by multiple selectors for a single state change.

The biggest reason that the performance increase is so big is pretty much that the major browsers and babel haven't gotten around to optimizing spread operators right now so selectors as they are written in reselect v2 are using a lot of cycles to copy memory and then GC that memory for every time they are called.

We're using a fork of reselect which includes this PR in GitKraken which dramatically speeds up the application. Our particular use case is that we use reselect for almost all projections of the state and in many react-redux container components. The breadth and depth of our selector trees before this PR would cause smaller actions to become noticeably slower and also GC thrashing from all of the extra superfluous data created from the cache hit calculations.

I can try to provide some profiler data if you need more numbers. And thanks for an amazing library in reselect. It's been really helpful for me and my team for making our application's code and architecture better.

src/index.js
```diff
@@ -2,19 +2,38 @@ function defaultEqualityCheck(a, b) {
...
-    if (
-      lastArgs !== null &&
-      lastArgs.length === args.length &&
-      args.every((value, index) => equalityCheck(value, lastArgs[index], index))
-    ) {
-      return lastResult
+
+  // we reference arguments instead of spreading them for performance reasons
+  return function () {
+    if (!areArgumentsShallowlyEqual(equalityCheck, lastArgs, arguments)) {
```
philpl: While unintuitive, leaking the "arguments" is actually listed as an optimization killer here: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

Can you benchmark what the performance looks like if you don't leak it? Thanks, I'd love to see that. If you're short on time, I'd look into that myself.

philpl: It could actually be that v8 is going to Inline the comparison, in which case this doesn't matter, except for leaking it into the computation function itself

johnhaley81: The only work-around (that I know of) for referencing arguments in a closure scope is to copy those arguments which is what the original code was doing. Ironically, removing the code that allowed chrome to optimize those functions dramatically sped up the code. I'll whip up a quick example and post it to another branch for you.

johnhaley81: @philpl https://github.com/johnhaley81/reselect/tree/add-arguments-v8-optimization

johnhaley81: Local perf tests show:

 - Post V8 args optimization: ~350ms
 - Pre V8 args optimization: ~200ms

alex3165: Found this jsperf if it can help: https://jsperf.com/rest-vs-arguments

johnhaley81:
So coveralls thinks the coverage has gone down and codecov thinks it's the same... That's what I get for adding a test :/

johnhaley81:
This is GTG again. Sorry about all of the coverage spam.

ellbee:
Ok, I'm convinced. Thanks for your thoroughness @johnhaley81, and sorry it took so long to get to!

johnhaley81:
Woot!!

src/index.js
philpl: Lgtm! Thanks for checking that this has a negative impact here

gaearon:
Good stuff.

sdtsui:
Hooray!

PR #195, opened 2016-11-24
Adds test ensuring defaultMemoize passes correct params to equalityCheck

jimbolla:
This test is to prevent a regression that would occur in PR #193 by demonstrating why it's important that `lastArgs` is assigned each time.


ellbee:
Thanks @jimbolla

johnhaley81:
Thanks @jimbolla that really helps explain what you were talking about in #193.

PR #197, opened 2016-11-30
Fix minor typo in README.md

carlbernardo:
Due to non-vowel sound of hypothetical, it should be 'a' instead of 'an'

https://en.oxforddictionaries.com/definition/hypothetical


ellbee:
Thanks!

PR #199, opened 2016-12-09
Update eslint and eslint-plugin-react

wbreakell:
This updates eslint and eslint-plugin-react dev dependencies. After updating these dependencies, eslint started reporting a `no-unused-vars` error in test/test_selector.js:
```
70:9  error  'called' is assigned a value but never used  no-unused-vars
```
I've removed the unused variable and eslint now succeeds.


wbreakell:
It looks like ESLint is failing when tested against Node.js v0.12. ESLint v3 drops support for Node.js versions prior to 4 http://eslint.org/docs/user-guide/migrating-to-3.0.0#dropping-support-for-nodejs--4

PR #201, opened 2016-12-13
Allow importing of reselect flow types

johnhaley81:
By putting the `Selector` types into the `declare module 'reselect'` block you can let users of the library import the types of the library using `import type { Selector } from 'reselect'` which is very helpful for users of flow.

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -1,675 +1,675 @@
-type Selector<TState, TProps, TResult> = {
-    (state: TState, props: TProps, ...rest: any[]): TResult;
-};
-
-type SelectorCreator = {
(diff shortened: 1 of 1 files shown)


ellbee:
Does anyone know what the current recommended practice is for Flow typings? Reselect has typings in Flow Typed, so should we maybe remove the typings from this repo altogether?

johnhaley81:
👋 

It did confuse me a little that the flow types were in the library since it does seem like most flow typings are in https://github.com/flowtype/flow-typed.

I'll close this PR since it does break CI and I haven't had time to dig into why that is. Also it does seem like the `Selector` type does not need to be exported since it should just be created under the hood.

Thanks @ellbee

PR #210, opened 2017-01-10
cacheSize option for defaultMemoize

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
As 3rd argument, now defaultMemoize() can get the `cacheSize`.

There are two arrays to store previous argument arrays, previous results.
```
let cacheArgsArr = []
let cacheREsultArr = []
```

`cacheSize` feature is designed as follows:

```
const createSelector = createSelectorCreator( defaultMemoize, void 0, 2 )
const powerSelector = createSelector([
      (state)=>state.num
], ( num )=> num*num )

console.log( powerSelector({ num: 1 }) )            // 1 (recomputation)
console.log( powerSelector.getCacheArgsArr() )      // [ [1] ]
console.log( powerSelector.getCacheResultArr() )    // [1]

console.log( powerSelector({ num: 2 }) )            // 4 (recomputation)
console.log( powerSelector.getCacheArgsArr() )      // [ [2],[1] ]
console.log( powerSelector.getCacheResultArr() )    // [ 4, 1 ]

console.log( powerSelector({ num: 3 }) )            // 9 (recomputation)
console.log( powerSelector.getCacheArgsArr() )      // [ [3],[2] ]
console.log( powerSelector.getCacheResultArr() )    // [ 9, 4 ]

console.log( powerSelector({ num: 2 }) )            // 4 (cache)
console.log( powerSelector.getCacheArgsArr() )      // [ [2],[3] ]
console.log( powerSelector.getCacheResultArr() )    // [ 4, 9 ]
```

In general, there is high a probability that last argument is used again.
So last arguemnts and result are added or updated to index: 0 in cacheArgsArr` and `cacheResultArr`

Please feel free to contact us if you have something lacking.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,22 +2,67 @@ function defaultEqualityCheck(a, b) {
   return a === b
 }
 
-export function defaultMemoize(func, equalityCheck = defaultEqualityCheck) {
-  let lastArgs = null
-  let lastResult = null
-  const isEqualToLastArg = (value, index) => equalityCheck(value, lastArgs[index])
-  return (...args) => {
-    if (
-      lastArgs === null ||
-      lastArgs.length !== args.length ||
-      !args.every(isEqualToLastArg)
-    ) {
-      lastResult = func(...args)
+export function defaultMemoize(func, equalityCheck = defaultEqualityCheck, size=1) {
+
+  let cacheArgsArr = []     // Cache store of old argument arrays
+  let cacheResultArr = []  // Cache store of old results
+
+  const memoizedResultFunc = (...args) => {
+    let foundIdx = -1
+    const oldCacheResultArr = cacheResultArr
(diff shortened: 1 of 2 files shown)


ivawzh:
Thank @b6pzeusbc54tvhw5jgpyw8pwz2x6gs , this was an enlightenment.
What do you think about the already supported selector customization below:

```js
import memoizer from 'fast-memoize'
import { createSelectorCreator } from 'reselect'

// Why not Lodash's memoize?
// Because that won't work when the function takes more than one argument. And this is fast.

export const createSeletor = createSelectorCreator(memoizer)
```

adnelson:
@ivawzh that's a great solution -- it would be super helpful if this were in the `README.md`. It suggests to "Create a custom selector with a cache size greater than one." But doesn't provide any guidance on how to do so.

PR #213, opened 2017-01-19
A new approach to sharing selectors: createIndexedSelector

1000hz:
I think there’s a better approach to sharing selectors with props between components than the one outlined in the docs. That approach suggests each component should create its own private instance of the selector, but this leads to unnecessary recomputations. Two different selector instances given the same arguments will individually recompute and store the same result.

Instead of creating a private instance of the selector for each component, we could instead create a higher-order selector that keeps a cache of selectors and instantiates a new instance only when necessary. We could call this concept an Indexed Selector.

```js
const indexResolver = (state, index) => index

export function createIndexedSelector(selectorFactory, resolver = indexResolver) {
  const cache = {}
  return (...args) => {
    const key = resolver(...args)
    const selector = cache[key] || (cache[key] = selectorFactory())
    return selector(...args)
  }
}
```

Note that for convenience and generalizability, I’ve opted to pass the index to selectors instead of a props object by default, but you can use your own resolver if you'd rather pass in your whole props object to your selectors. Rewriting the docs example to use this pattern, we get:

```js
import { createSelector, createIndexedSelector } from 'reselect'

const getVisibilityFilter = (state, index) =>
  state.todoLists[index].visibilityFilter

const getTodos = (state, index) =>
  state.todoLists[index].todos

const getVisibleTodos = createIndexedSelector(() => {
  return createSelector(
    [ getVisibilityFilter, getTodos ],
    (visibilityFilter, todos) => {
      switch (visibilityFilter) {
        case 'SHOW_COMPLETED':
          return todos.filter(todo => todo.completed)
        case 'SHOW_ACTIVE':
          return todos.filter(todo => !todo.completed)
        default:
          return todos
      }
    }
  )
})

export default getVisibleTodos
```
```js
const mapStateToProps = (state, props) => {
  return {
    todos: getVisibleTodos(state, props.listId)
  }
}
```

This has the added bonus of simplifying our connector since we no longer need to use `makeMapStateToProps`!

Let me know of anything I might be overlooking or further changes you’d like to discuss.


walerun:
It looks interesting. But what if components have unique key/id each time when we create them. If we will delete some component, it means cache will have an unused selector, wich will never invoke. It may cause cache growth. Though maybe it's not a big problem.
I have one more question. Maybe it is not related to current PR, but how can we combine components selectors to create a new selector?

toomuchdesign:
Hi all! I've just published a tiny plugin to memoize a collection of selectors based on a resolver function for a project of mines.

It works more or less as Lodash's .memoize() does, but with almost zero configuration. The name is [re-reselect](https://www.npmjs.com/package/re-reselect) and it's just a few line of code without dependencies.

If externalizing to another micro-package is ok, It should fit this multi-component scenario, too.

Ping me for any info.
Thank's for the great work you're doing on Reselect. Cheers!

drcmda:
@ellbee @toomuchdesign 

Could something like createCachedSelector be officially taken in? I think it's a really nice idea and it would go a long way in easing the complexity. 

Our team is kind of scared committing to reselect because in essence it would force us to use the factory pattern pretty much all the time because properties in Redux always change while the state often remains the same. 

We feel like the smallest use-case, one that rarely ever applies to a real world scenario, has been made default while in the majority of cases mapStateToProps has to be padded with boilerplate which offloads the cognitive load to the component designer that now has to think through intricacies that aren't of his/her concern at all. The designer now doesn't just pick state, but reselect itself has to be taken into account, how it works, how it memoizes, if the component selects slices of state according to props or if it's static, etc.

toomuchdesign:
@drcmda 

Your point is the reason why I published re-reselect. I just moved a memoized reselect factory into a separate package declaring reselect as peerDependency.

ellbee:
I think what I'll probably do here is advertise re-reselect in the documentation, and if it turns out to be very popular then we will pull it into reselect itself.

bilby91:
I'm using `re-reselect` now and was a smooth transition with minor adjustments to my code base. Works really good! Thanks @toomuchdesign !

PR #226, opened 2017-03-15
Test to ensure inputSelectors called with same params as the selector

AugustinLF:
On master, the inputSelectors are called with an extra `undefined` param when the selector is only called with the state. This is usually not a problem, but in my specific case, some inputSelectors are curried functions, and in the result func, I use the returned function. I know it kind of defies the concept of `createSelector`, but in these cases, perf does not matter, and it lets me use the same API/patterns consistently in my app.

This bug is due to the `props` param (possibly undefined) passed to `dependency`, on this [line](https://github.com/reactjs/reselect/blob/master/src/index.js#L54).

I was planning to submit a fix, however, the 3.0.0 rewrite did fix this bug! So here is a test to prevent any regressions on this side :) If you add this test to master, you'll see that it breaks.

I can also submit this (with the corresponding fix) to the current master branch, if you think 3.0.0 won't be publish soon. I have no emergency on this, I can use a simple workaround on my side, it just makes my code a bit less readable!

### DIFF
--- a/test/test_selector.js
+++ b/test/test_selector.js
@@ -23,7 +23,7 @@ suite('selector', () => {
     const firstState = { a: 1 }
     const firstStateNewPointer = { a: 1 }
     const secondState = { a: 2 }
-    
+
     assert.equal(selector(firstState), 1)
     assert.equal(selector(firstState), 1)
     assert.equal(selector.recomputations(), 1)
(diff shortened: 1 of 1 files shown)


ellbee:
I've published 3.0.0 now, so I guess we can close this? (Thanks very much for making the PR by the way!)

AugustinLF:
Awesome! I'll test the 3.0.0 version asap, to confirm that it does fix my issue!

However, I still do think that this test should be merged (I changed the base branch to master), since I would consider this behaviour as part of the "public API", and breaking this test could break things in some apps :)

ellbee:
Ah right, sorry I misunderstood the PR slightly. I'll go ahead and merge.

Thank you!

PR #227, opened 2017-03-16
Simplify `Sharing Selectors` example

danny-andrews:
In the `Sharing Selectors with Props Across Multiple Components` we change the selector to be a selector factory **and** use the function parameter of mapStateToProps, when only the former is necessary to provide each container instance with its own selector. [Proof](http://codepen.io/danny-andrews/pen/WpXRvZ?editors=0111).

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,25 +352,7 @@ const makeGetVisibleTodos = () => {
 export default makeGetVisibleTodos
 ```
 
-We also need a way to give each instance of a container access to its own private selector. The `mapStateToProps` argument of `connect` can help with this.
-
-**If the `mapStateToProps` argument supplied to `connect` returns a function instead of an object, it will be used to create an individual `mapStateToProps` function for each instance of the container.**
-
-In the example below `makeMapStateToProps` creates a new `getVisibleTodos` selector, and returns a `mapStateToProps` function that has exclusive access to the new selector:
-
-```js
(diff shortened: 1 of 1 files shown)


markerikson:
Uh... no, no, that's _not_ what this example is trying to show.  What you've changed it to is a "basic" example of `mapState`.  Yes, that syntax works, but the example is trying to show something else instead.

Actually, looking at it, your change definitely defeats the purpose.  The point is create a unique instance of the selector _for that component instance_, so that the selector's caching is based on _this_ component's data, not shared across all components of that type.  By creating the selector instance _inside_ of `mapState`, not only does it not do caching for that component instance, it doesn't do _any_ caching at all because it's creating a new selector instance _every_ time `mapState` is called.

I appreciate the time you took to file the PR, but this is not correct.  :(

danny-andrews:
Thanks for the explanation @markerikson. I had a suspicion something was wrong with this, but it appeared to work. Looking at it now, I see there error.

markerikson:
Sure.  Hopefully my explanation makes sense.  I also wrote an explanation of `map*` factories and selector creation a while back in a Reddit comment: https://www.reddit.com/r/reactjs/comments/5dxasp/any_deepdiveadvanced_tutorials_on_reselect/ .

Again, definitely don't want to discourage contributions, but I _do_ have to help make sure they're correct :)

PR #238, opened 2017-03-29
Added a memoize function with a configurable cache size.

ndbroadbent:
I was using `memoize` from `lodash` to create an unbounded cache, which is fine for some parts of my code. I also know about `re-reselect`, but that solves a slightly different set of problems.

There are a few cases in my app where I really, really want to have a specific cache size, so that it would store the last 2 or 3 results. One example is where I am using selectors for both old and new props, and comparing the results. Another example is where I have a selector that depends on a frequently changing boolean value. Instead of recalculating every time, I just want to store both results, then throw them both away if something else changes. I ended up writing a "switching" selector creator:

```
const createSwitchingSelector = (inputSelectors, switchSelector, resultFunc) => {
  const falsySelector = createSelector([...inputSelectors, () => null], resultFunc)
  const truthySelector = createSelector([...inputSelectors, switchSelector], resultFunc)
  return state => (
    switchSelector(state) ? falsySelector(state) : truthySelector(state))
}
```

But I kept running into places where I really want better control over caching, so I've created a new memoization function that accepts a `cacheSize` argument. It sets up arrays for the cached args and results, and then cycles through the array, overwriting the oldest results. It also starts searching at the newest index.


ndbroadbent:
I made some changes recently:

I added a `lastCacheHitIndex` variable, so that any subsequent lookups do not have to loop through the equality check. If the result is not found at `lastCacheHitIndex`, it will begin searching from newest to oldest.

After running the performance benchmarks many times, I concluded that the new `defaultMemoize` performance is just as good as the original one, so there is no need to keep two versions. Also, after removing the old version, I realized that I had broken the *"recomputes result after exception"* and *"memoizes previous result before exception"* tests, because my new code was setting the args and results in the wrong order. So it's far better to just keep one version of the `defaultMemoize` function, to catch bugs in one place.

ndbroadbent:
Also it would be great if coveralls could only comment when there's a change

ndbroadbent:
cc: @ellbee

I was wondering if someone might be able to take a look at this? I've been using this branch in my own app for the last few weeks, and it seems to be working great for me.

b6pzeusbc54tvhw5jgpyw8pwz2x6gs:
@ndbroadbent 
On March 25th they added [re-reselect](https://github.com/toomuchdesign/re-reselect) to the README. I think they are trying to solve the cache function through `re-reselect`.

ndbroadbent:
Hi @b6pzeusbc54tvhw5jgpyw8pwz2x6gs - I actually use `re-reselect` as well, but I found that I needed both types of caching. The problem with `re-reselect` is that the cache is unbounded, so in my app that would cause a bad memory leak. There are some cases where it's really nice to be able to set a fixed cache of 2 or 3, and then you don't have to worry about it.

timhwang21:
@ellbee any possibility of an update on this? I have a use case where I need a cache size of exactly two. This would be extremely handy. Thanks!

toomuchdesign:
Ping 😃.

From version **1.0.0**, `re-reselect` comes with a [**cacheObject** option](https://github.com/toomuchdesign/re-reselect#optionscacheobject) to provide custom caching behaviours. The library expose also **3 pre-configured cache strategies** which might comes in handy in @ndbroadbent case:
- flat
- fifo
- lru

Greetings!

Issue #237, opened 2017-03-28
Type error when using created selectors as input for a new selector.


kevinwardenga:
After the upgrade to version 3.0.0 the typescript compiler complains about a non matching argument type. Using the following code:

```ts
export const selectProductStore = (state: any): State => state.productStore;
export const getEntity = createSelector(selectProductStore, store => store.entity);
export const getTitle = createSelector(getEntity, entity => entity.title);
```

Is producing the following error:
```
Argument of type 'OutputSelector<any, Product, (res: State) => Product>' is not assignable to parameter of type '[ParametricSelector<{}, {}, {}>, ...'. 
Property '0' is missing in type  'OutputSelector<any, Product, (res: State) => Product>'
```

The same code doesn't show any typing error with version 2.5.4.

alex3165:
I noticed it as well, would be nice to add your example to https://github.com/reactjs/reselect/blob/master/typescript_test/test.ts so that we handle this use case in the typescript test and fix it. I think that is because `Selector` is not the same type as `OutputSelector` so one fix would be to only use `OutputSelector` everywhere and put `resultFunc `, `recomputations ` and `resetRecomputations ` Optional, @aikoven what do you think about it ?

alex3165:
I tried to reproduce your issue but didn't get any error : https://github.com/alex3165/reselect/commit/388da067605a6a9ba051c5784111ea2b7b3265f7#diff-2fe2e76dc80d303b543f572094baf958R35

kevinwardenga:
The strange thing is the code compiles fine and runs without errors, just the editor is showing the error and therefore autocompletion is broken. I tried sublime and atom both showing this error when hovering over `getEntity`.

alex3165:
@kevinwardenga It is probably because your IDE is using an old version of Typescript

kevinwardenga:
I've already checked that, it's using the latest typescript version (2.2.2).

aikoven:
> After changing Selector and ParametricSelector from type to an interface the errors are gone and everything is working as expected.

As far as I remember, extending type alias was an error before. Although, latest TypeScript doesn't give me any error. Maybe your editor uses some built-in TypeScript compiler that is outdated.

Either way, I've created a PR https://github.com/reactjs/reselect/pull/240 that replaces type aliases with interfaces.

kevinwardenga:
[...]

PR #240, opened 2017-03-30
Update Selector types for correct extension

aikoven:
Fixes https://github.com/reactjs/reselect/issues/237.
This is backwards-compatible.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,18 +1,16 @@
 export as namespace Reselect;
 
-export interface Selector<S, R> {
-  (state: S): R;
-}
-export interface OutputSelector<S, R, C> extends Selector<S, R> {
+export type Selector<S, R> = (state: S) => R;
+
+export type OutputSelector<S, R, C> = Selector<S, R> & {
   resultFunc: C;
   recomputations: () => number;
   resetRecomputations: () => number;
 }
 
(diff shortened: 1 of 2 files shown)


aikoven:
I've removed interface extension completely to address https://github.com/reactjs/reselect/issues/241

aikoven:
This should remain backward-compatible because the last commit could only potentially break TS 2.1 users who extend `OutputSelector` interface, but TS 2.1 support is already broken in 3.0.0 (see #237).

For TS 2.2 extending intersection type will work fine.

aikoven:
I realized that I didn't add a test for this, please don't merge yet.

aikoven:
Done.

danielduhh:
when will this be published ?

aikoven:
cc. @ellbee

alex3165:
I think we can go ahead with that

the-destro:
@ellbee et. al: can we get a 3.0.1 release with this fixed please?

richarddavenport:
Please merge!

ellbee:
@alex3165 Feel free to release and publish if you think this is good to go.

alex3165:
Merged, released and published with the patch version `3.0.1`

ellbee:
Thanks @alex3165!

Issue #162, opened 2016-09-01
reconsider allowing selectors to take arguments


jwoudenberg:
So I read the issues related to selectors taking no arguments (#18, #38, #59) and I understand the reasoning behind it: that if you feel the need to create a selector with a dynamic argument, probably that argument should be in your store state somewhere.

Only I don't believe that's necessarily true. In a project I'm working on we keep the router state in the router itself and don't jump through extra hoops to sync this with the store state. I think this too is [a pretty well established pattern](http://stackoverflow.com/questions/36722584/how-to-sync-redux-state-and-url-hash-tag-params/36749963#36749963). It would be great to be able to pass the router state into a selector as an argument.

I'm curious to hear your thoughts :smile:.

ellbee:
When using `connect`, component props are passed as the second argument to the selector. This allows you to use the pattern referenced in your link.

More details here: https://github.com/reactjs/reselect#createselectorinputselectors--inputselectors-resultfunc

jwoudenberg:
I should have found that myself, that works nicely. Thanks for your reponse!

CharlieEarnup:
@ellbee what if you don't want to couple your component code to your selector code? What's to stop someone from passing in a custom props object there? Isn't the essentially the same as passing args?

PR #242, opened 2017-04-08
Add note about Redux props to FAQ

joshkel:
The FAQ states that Reselect doesn't take additional arguments, but this doesn't cover the common scenario of those additional arguments being available as Redux props. (I needed a selector that took some additional arguments (which were available as Redux props), went straight to the FAQ (overlooking the "example" section), and saw an answer that I thought meant that I was out of luck. #162 seems to be from a user who followed a similar thought process.) This change hopefully clarifies things and adds an additional place to access this information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -754,7 +754,9 @@ A: Yes. Reselect has no dependencies on any other package, so although it was de
 
 ### Q: How do I create a selector that takes an argument?
 
-A: Reselect doesn't have built-in support for creating selectors that accepts arguments, but here are some suggestions for implementing similar functionality...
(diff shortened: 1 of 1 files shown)


ellbee:
Great, thanks!

alexrussell:
I know this is merged and all, but should the wording not say "React props" instead of "Redux props" or am I getting the wrong end of the stick here?

joshkel:
@alexrussell - Oops. Thanks for catching that. I just submitted #244.

ellbee:
Argh, I missed it too! Thanks @alexrussell

PR #243, opened 2017-04-12
Fixed typings

rikkertkoppes:
This does not compile

    export type Selector<S, R> = (state: S) => R;
    export interface OutputSelector<S, R, C> extends Selector<S, R> {
      resultFunc: C;
      recomputations: () => number;
      resetRecomputations: () => number;
    }

interfaces cannot extend types.

https://github.com/reactjs/reselect/blob/master/src/index.d.ts#L3-L8

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,14 +1,14 @@
 export as namespace Reselect;
 
 export type Selector<S, R> = (state: S) => R;
-export interface OutputSelector<S, R, C> extends Selector<S, R> {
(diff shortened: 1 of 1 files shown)


aikoven:
Please see https://github.com/reactjs/reselect/pull/240

Also, you can switch to TypeScript 2.2 where interfaces extending type aliases is allowed.

alex3165:
This is fixed by https://github.com/reactjs/reselect/pull/240 I have just released it with `3.0.1`

PR #244, opened 2017-04-13
Fix terminology

joshkel:
"React props," not "Redux props." This fixes the terminology for my change in #242.


ellbee:
Thank you!

PR #249, opened 2017-04-24
mention cache size in the problem

stevemao:
I didn't understand why a selector created with `createSelector` doesn't simply return a cache on the same set of arguments until I read [Q: Can I share a selector across multiple components](https://github.com/reactjs/reselect#q-can-i-share-a-selector-across-multiple-components). I think mentioning cache size would help to understand it better.


ellbee:
Thanks

PR #263, opened 2017-06-26
Add minified umd package.

kaidjohnson:
First off, thanks for this library! It is the missing link between redux stores and view components and is absolutely fantastic!

Secondly, I realize unminifed, the library is like 4.6k, so this isn't a huge issue. I do think, though, as a best practice, all umd libs should provide both a minified and an unminified version of the code. As such, I've added uglify to compress the dist lib down to a minified version coming in at just over 2k.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,10 @@
   },
   "scripts": {
     "compile:commonjs": "better-npm-run compile:commonjs",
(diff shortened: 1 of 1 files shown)


ellbee:
Great, thank you!

PR #274, opened 2017-07-28
Allow Differently Typed Selectors

benshope:
Here is the related issue https://github.com/reactjs/reselect/issues/264

Right now we have to do typings like this:

```
const selectorOne = (state: State) => state.something;
const selectorTwo = (state: State) => state.other;
createSelector([selectorOne, selectorTwo], () => ...);
```

This PR allows us to do typings like this:

```
const selectorOne = (state: { something: string }) => state.something;
const selectorTwo = (state: { other: number }) => state.other;
createSelector([selectorOne, selectorTwo], () => ...);
```

Let me know if I should change anything, thanks!

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -953,8 +953,8 @@ export function createSelector<S1, S2, S3, S4, S5, S6, S7, S8, S9, S10, S11, S12
 ): OutputParametricSelector<S1 & S2 & S3 & S4 & S5 & S6 & S7 & S8 & S9 & S10 & S11 & S12, P1 & P2 & P3 & P4 & P5 & P6 & P7 & P8 & P9 & P10 & P11 & P12, T, (res1: R1, res2: R2, res3: R3, res4: R4, res5: R5, res6: R6, res7: R7, res8: R8, res9: R9, res10: R10, res11: R11, res12: R12) => T>;
 
 
-export function defaultMemoize<F extends Function>(
-  func: F, equalityCheck?: <T>(a: T, b: T, index: number) => boolean,
+export function defaultMemoize<F extends Function, T>(
+  func: F, equalityCheck?: (a: T, b: T, index: number) => boolean,
 ): F;
 
 export function createSelectorCreator(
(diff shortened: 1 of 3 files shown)


aikoven:
This is a breaking change since generic arguments changed. We should include both old and new signatures.

benshope:
@aikoven Yep!  Added them back

aikoven:
I've tested these against a fairly large project with lots of selectors and it worked fine.

There's a failing test that should be fixed. Also, I guess some tests for the new signatures should be added as well.

benshope:
@aikoven  Cool!  I fixed the broken tests and added both a failing-case and a passing-case test for both parametric and array selectors

benshope:
Also, I put the Python script I threw together into a Gist - in case someone wants to refer back to this and change the types at some point https://gist.github.com/benshope/9712ed6d725a263467447ac6473502a9

aikoven:
Nice work @benshope! It's 👍 to me, I hope someone else would also test these in their project.

seepel:
@aikoven I tacked on an additional commit to this PR to cover typescript 2.4's new strict generic checks and added a test  here: https://github.com/reactjs/reselect/pull/274/commits/5ec2b1f29bf7100420a93736cbe7996eb48a6687

You can see relevant discussion on the topic in this typescript issue: https://github.com/Microsoft/TypeScript/issues/17509

aikoven:
Changing the generic signature of `defaultMemoize` is a breaking change. 

It could potentially be solved via generic defaults, although it would break for users of TS 2.2 and below.

seepel:
That's fair, I hadn't considered the case where some one might be calling `defaultMemoize` with the generic:
```typescript
defualtMemoize<MyFunction>
```

I've reverted the commit.

benshope:
@aikoven Hey!  Any updates on the status of this?

benshope:
@aikoven Cool!  Yes these changes should be backwards compatible

ellbee:
Sorry to be so slow to this. As @aikoven has approved, I'll go ahead and commit. Thanks!

benshope:
@ellbee Thanks!!

PR #278, opened 2017-08-25
Update README

ericandrewlewis:
It's useful to note that selectors will be called every time the state tree is updated, rather than when the component is updated.


markerikson:
Thanks!

PR #283, opened 2017-09-18
added codepen link for first README example

Domiii:
Having a codepen to play around with right away, without having to waste 1-10 minutes (depending on experience) can be a blessing!


ellbee:
Cool, thanks! Sorry it took so long for me to get around to looking at this.

PR #288, opened 2017-10-03
Fix typo in README

extg:


extg:
I'm sorry, it was right

PR #291, opened 2017-10-05
chore: added `module` key to let modern bundlers pick ES version

bhovhannes:
`jsnext:main` is deprecated in favor of `module` key.
We can keep both to ensure the greatest compatibility with all bundlers.


ellbee:
Great, thanks @bhovhannes!

PR #297, opened 2017-11-11
Use provided memoize function for selectors

Mosho1:
In addition to memoizedResultFunc.


skortchmark9:
@ellbee could you take a look at this? It would be really handy to have in

ellbee:
Ok, this looks good to me. Will be released in version 4. Thanks!

PR #302, opened 2017-11-27
Updated reselect tools verbiage

skortchmark9:
Reflect that the chrome extension is out and be more specific about what it can do!


ellbee:
Thanks!

PR #316, opened 2018-01-21
Fix grammatical error

gitname:
The original verb, "change," does not agree with the subject, "one" (as opposed to, "arguments").

* Reference: Guideline number 5 on [Purdue University > ... > Grammar > Subject/Verb Agreement](https://owl.english.purdue.edu/owl/resource/599/01/)


markerikson:
Thanks!

PR #321, opened 2018-02-09
Allow selector dependencies to be selector factories

madcapnmckay:
As noted in the docs, if you are not careful, memoization can be undone when using a selector with props in multiple components.

The solution is to use a selector factory and pass that to mapStateToProps instead. This works well for one level deep selectors, however, if we are composing memoized selectors the same caveats apply. It is very easy to unintentionally be undoing all the memoization as we nest. The solution again is to keep wrapping our [sub-selectors in factories](http://jsbin.com/bohojakobo/1/edit?js,console) and executing them as we create the selectors.

A neater approach is to always accept either a selector or a selector factory in the createSelector function and handle it in a similar manner as react-redux connect does. This would allow us nest selector factories without having to explicitly execute the factory.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,10 +46,6 @@ function getDependencies(funcs) {
     )
   }
 
-  // return dependencies.map(dep => {
-  //   console.log('mapping dep ' + dep.length + ' ' + dep.toString());
-  //   return dep.length === 0 ? dep() : dep;
-  // })
   return dependencies
 }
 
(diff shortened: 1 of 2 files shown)


toomuchdesign:
Hi @madcapnmckay, just to point out that [`re-reselect`](https://github.com/toomuchdesign/re-reselect), let you declare selectors which **internally behave like a factory** but keep the same **API of a normal `reselect` selector outside**.

This allows [forgetting differences between **selectors** and **selector factories**](https://github.com/toomuchdesign/re-reselect/blob/master/examples/2-avoid-selector-factories.md).

abritinthebay:
@toomuchdesign while true, that's somewhat different to the proposed solution.

PR #331, opened 2018-03-16
Fix error in README example

hwaterke:


markerikson:
The example is actually correct, because `connect` will use the returned "real" `mapState` function instead.

PR #332, opened 2018-03-19
Update README.md

didierfranc:


ellbee:
Thanks! As pointed out in #336 I think we can go further with this, but as a first small step this is good.

PR #220, opened 2017-02-24
Expose custom-memoize function - ability to clear cache per selector

schneckentempo:
I found, that lodash.memoizing selectors with bigger datasets can lead to quite a large heap, therefor I want to delete the cache at a certain point of the application.
PR regarding this issue/feature-request #219

PR #358, opened 2018-07-18
add memoizedResultFunc to output selector interface

montemishkin:
Exposes the `memoizedResultFunc` on the output selector so that (for example) you can access the cache (for example to clear it) when using `lodash.memoize` as is done in [this example](https://github.com/reduxjs/reselect#use-memoize-function-from-lodash-for-an-unbounded-cache).

I see that https://github.com/reduxjs/reselect/pull/220 exists, but it renames the exposed field to `memoize` which is rather misleading.  Also, it hasn't been touched for a while, so i dont know what the status is there.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -3,6 +3,7 @@ export as namespace Reselect;
 export type Selector<S, R> = (state: S) => R;
 
 export type OutputSelector<S, R, C> = Selector<S, R> & {
+  memoizedResultFunc: C;
   resultFunc: C;
   recomputations: () => number;
   resetRecomputations: () => number;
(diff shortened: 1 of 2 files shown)


montemishkin:
howdy @mnorrish I was hoping you could take a look at this (since you stamped #220)

My thought is that we should land this, and cancel #220.

Let me know what you think / if there is somebody else we can direct this too :)

montemishkin:
howdy @ellbee I was hoping you could merge this (I didn't realize this repo has write access restrictions).  

Let me know what you think / if there is somebody else I can direct this to :)

PR #359, opened 2018-07-20
Use provided memoize options in outer memoize call

montemishkin:
There are two memoize calls in `createSelectorCreator`.  One is to memoize the so called `resultFunc`.  The other memoize allows us to skip out early if the resulting selector is called with the exact same arguments.  

Currently, this second memoize uses the caller's provided `memoize` implementation, but does not pass the caller's `memoizeOptions`.  This PR changes it to pass the `memoizeOptions`.

### DIFF
--- a/test/test_selector.js
+++ b/test/test_selector.js
@@ -247,10 +247,12 @@ suite('selector', () => {
       a => a
     )
     assert.equal(selector({ a: 1 }), 1)
-    assert.equal(selector({ a: 2 }), 1) // yes, really true
+    // still returns `1` since `typeof state` is still "object"
+    assert.equal(selector({ a: 2 }), 1)
     assert.equal(selector.recomputations(), 1)
-    assert.equal(selector({ a: 'A' }), 'A')
-    assert.equal(selector.recomputations(), 2)
+    // again returns `1` since `typeof state` is still "object"
+    assert.equal(selector({ a: 'A' }), 1)
+    assert.equal(selector.recomputations(), 1)
   })
   test('custom memoize', () => {
     const hashFn = (...args) => args.reduce((acc, val) => acc + '-' + JSON.stringify(val))
--- a/src/index.js
+++ b/src/index.js
@@ -65,18 +65,21 @@ export function createSelectorCreator(memoize, ...memoizeOptions) {
     )
 
     // If a selector is called with the exact same arguments we don't need to traverse our dependencies again.
-    const selector = memoize(function () {
-      const params = []
-      const length = dependencies.length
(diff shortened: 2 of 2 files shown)


test/test_selector.js
```diff
@@ -247,10 +247,12 @@ suite('selector', () => {
       a => a
     )
     assert.equal(selector({ a: 1 }), 1)
-    assert.equal(selector({ a: 2 }), 1) // yes, really true
+    // still returns `1` since `typeof state` is still "object"
+    assert.equal(selector({ a: 2 }), 1)
+    assert.equal(selector.recomputations(), 1)
+    // again returns `1` since `typeof state` is still "object"
+    assert.equal(selector({ a: 'A' }), 1)
```
montemishkin: its worth highlighting that this is a subtle change in behavior and could break dependent codebases.  So bumping a version number seems to be in order

PavelDemyanenko:
Any news on this request? Looks good to me

Zhuvikin:
Faced the same issue. Is it planned to be merged?

ms88privat:
I just investigated a related topic. I really think we should pass two memoize and two options to this selectorCreator, so everybody can decide what they want exactly. I was not happy when https://github.com/reduxjs/reselect/pull/297 was introduced. This forced me to do an expensive `isEqual` check on the broad input instead of the very lightweight outcome for the input of the result function. @ellbee What do you think about that?

yoavniran:
facing the same issue exactly. passing memoizeOptions to the sub selectors makes a lot of sense, or at least being able to control this with a flag so it can be opt it

oasisvali:
For those looking for a workaround in the meantime, I've found using the following pattern to be helpful:
```
// second argument only exists when the inner output function is memoized
const uniformMemoize = (func) => _.memoize(func, hashFn);

const createUniformMemoizeSelector = createSelectorCreator(
  uniformMemoize,
);
```

A nice side-effect of using this is that it unlocks the ability to use different memoization strategies for the overall selector arguments vs output function arguments. e.g.
```
// second argument only exists when the output function is memoized
const differentMemoize = (func, isOutput) => {
  if (isOutput) { return _.memoize(func, hashFn) }
  return defaultMemoize(func);
}

// this creates selectors with overall selector argument cache = defaultMemoize (cache size 1)
// output function argument cache = _.memoize w/ hashFn (size unlimited)
const differentMemoizeSelector = createSelectorCreator(
  differentMemoize,
  true,
);
```

markerikson:
v5 allows separate `memoize` and `argsMemoize` customization options

PR #381, opened 2018-10-17
Upgrade to babel@7

Andarist:


package.json
```diff
@@ -47,13 +47,7 @@
     "compile:commonjs": {
       "command": "babel -d lib/ src/ && ncp ./src/index.d.ts ./lib/index.d.ts",
       "env": {
-        "NODE_ENV": "commonjs"
-      }
-    },
-    "compile:umd": {
```
Andarist: personally I would move all builds (umd, umd:min, cjs, es) to rollup, but not sure how do u feel about it

markerikson:
Seems like this still is relevant.  @Andarist , any chance you could update the PR to merge cleanly?

Andarist:
@markerikson I've rebased and confirmed that it still compiles but please re-check the current setup.

Andarist:
Oh, I haven't spotted in a rush that this repo is not using Travis any longer 😅

markerikson:
that's because I only merged the other PR like 10 minutes ago :)

markerikson:
Well, this seems to build cleanly.  Let's get it in and move things along.  Thanks, @andarist !

PR #393, opened 2019-01-19
Updating And Improvement Typescript Infrastructure

sgrishchenko:
Feature scope:
* dependencies from #251 added in typings
* typescript and typings-tester versions updated (and tests fixed after that)
* improved type checking (`"strict": true` in tsconfig.json)
* added code generator for boilerplate code of `createSelector` overloads
* added formatting of generated typings with prettier (see `compile:typescript` in package.json)
* added type checking of generated typings with `tsc --noEmit`
* removed homogeneous selectors (heterogeneous selectors completely cover all cases)
* CI workflow changed (now types are being tested after compilation)

Minor changes:
* dropped CI build for Node v0.12 (for code generation used [Array.prototype.fill](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/fill), which availabale only in Node > v4)
* added CI build for Node v8


simperreault:
Any idea when this will be merged ? I'm really looking forward using the typing in my typescript unit tests.

anilanar:
👎  Those types obviously won't work. Parametric selectors are **never** uniform.

sgrishchenko:
@anilanar what did you mean? Unirorm it is homogenios or geterogenios? Why will not this typings working? There is special typing test, that checks typings? If you have broken code, can you send code snippet to check it?

anilanar:
Sorry I was hasty when I read the code, those types are pretty good.

sgrishchenko:
Are you saying about this `...args: any[]` in parametric selector?

anilanar:
Yes, exactly.

sgrishchenko:
Ou, I can not say that it is good solution for this library maintainers, but I think it is not linked with main goal off this PR. I just wanted to add dependecies in typings but when I tried to do this I was surprised how difficult to do this. So, I tried to simplify typings maintaining. But almost an year has passed and there's no activities from maintainers. They could say that they don't like solution with typings generation, I can prepare PR with hargcoded dependencies in typings. But if we are speaking about any parameters, I left them as is, because removing of them is breaking changes. But i didn't want to make breking changes, i wanted to make some small improvement. If you want to suggest removing of this any parameters, you can make separete PR with realisation of this suggestion. But thanks for activity in this PR, may be it will attract attention of maintainers.

anilanar:
@sgrishchenko If you publish a fork, I'd gladly use your typings.

PR #394, opened 2019-01-21
Fix createStructuredSelector() example snippet

dharkness:
The `createSelector()` version of the `createStructuredSelector()` example uses a third, undeclared input-selector. This pull request removes `mySelectorC` and `c` which don't appear in the second snippet.


markerikson:
Thanks!

PR #399, opened 2019-03-13
TypeScript: fix issue with `createSelector` overload order

OliverJAsh:
See the test I added which was failing before this change.

I can't decide whether or not the overloads should have worked in their previous order. I've filed a TS bug here just to check: https://github.com/Microsoft/TypeScript/issues/30369.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -29,16 +29,16 @@ export function createSelector<S, P, R1, T>(
 ): OutputParametricSelector<S, P, T, (res: R1) => T>;
 
 /* two selectors */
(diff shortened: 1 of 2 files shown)


OliverJAsh:
Actually I don't believe this is a bug with the type definitions because this change introduces a new bug (which there are currently no tests for)

``` ts
  type State = {foo: string};

    (state: State) => state.foo,
    (state: State) => state.foo,
    (foo, foo2) => foo,
  );

  // typings:expect-error
  selector2({foo: 'bar'}, {prop: 'value'});
```

Therefore I assume it's a TS bug, and I'll close this.

PR #410, opened 2019-05-21
remove the condition of next === null in the function areArgumentsShallowlyEqual

xiaoxiaojx:
I don't think it's meaningful to judge next=== null here. What's passed here is arguments!== null.


xiaoxiaojx:
prev=== null, which is used to determine whether it is the first call, next is fixed argument!== null

PR #417, opened 2019-07-20
we need the return of the makeXXX function as mapStateToProps, not th…

rangerch:
we need the return of the makeXXX function as mapStateToProps, not the function definition.


markerikson:
Sorry, that's not correct.  The "factory function" itself must get passed to `connect()`.

See another explanation of it here:

https://blog.isquaredsoftware.com/2017/12/idiomatic-redux-using-reselect-selectors/#advanced-optimizations-with-react-redux

PR #443, opened 2020-02-13
Add PURE annotation to a top-level createSelectorCreator call

Andarist:
This improves tree-shakeability of the `createSelectorCreator`


markerikson:
Seems reasonable to me, although I don't actually maintain Reselect.  @ellbee ?

Issue #428, opened 2019-10-19
No overload matches this call error for createStructuredSelector with typescript.


vijhhh2:
Hi, I am trying to build an react redux typescript app.

I am using reslect library to create selectors.

I am getting an error while using createStructuredSlectore. Please find the error beolow.

Error: No overload matches this call

Please find the code below:
```
import React from 'react';
import { connect } from 'react-redux';
import { createStructuredSelector } from 'reselect';

import { toggleDropDown } from '../../redux/cart/cart.actions';
import { selectItemCount } from '../../redux/cart/cart.selectors';

import { ReactComponent as ShoppingIcon } from '../../assets/11.2 shopping-bag.svg.svg';

import './cart-icon.component.scss';

interface ICartIconProps {
    toggleDropDown: () => void;
    itemCount: number;
}

const CartIcon: React.FC<ICartIconProps> = ({toggleDropDown, itemCount}) => (
    <div className='cart-icon' onClick={toggleDropDown}>
        <ShoppingIcon className='shopping-icon'/>
        <span className='item-count'>{itemCount}</span>
    </div>
)

const mapDispatchToProps = (dispatch: import('redux').Dispatch) => ({
    toggleDropDown: () => dispatch(toggleDropDown())
})

// If I use Below code its working fine
// const mapStateToProps = (state: AppState) => ({
//     itemCount: selectItemCount(state)
// })

// Iam getiing error here with below code
const mapStateToProps = createStructuredSelector({
    itemCount: selectItemCount,
})



export default connect(mapStateToProps, mapDispatchToProps)(CartIcon);
```
Slectors.ts
```
import { createSelector } from 'reselect';

const cart = state => state.cart;

export const cartItems = createSelector([cart], (cart) => cart.cartItems);

export const itemCount = createSelector([cartItems], (cartItems) => {
    return cartItems.reduce((accumulatedValue, cartItem) => accumulatedValue + cartItem.quantity, 0);
});

export const hidden = createSelector([cart], (cart) => cart.hidden);
```

cart.reducer.ts
```
import { CartActionsTypes } from "./cart.types";
import { addItemToCart } from "./cart.utils";

const INITIAL_STATE = {
    hidden:  true,
    cartItems: []
};

const cartReducer = (state = INITIAL_STATE, action) => {
    switch (action.type) {
        case CartActionsTypes.TOGGLE_CART_DROPDOWN:
            return {
                ...state,
                hidden: !state.hidden
            }
[...]

PR #454, opened 2020-05-02
createStructuredSelector type definitions with auto-infer

toomuchdesign:
### What kind of change does this PR introduce? _(Bug fix, feature, docs update, ...)_

Typing feature

### What is the current behaviour? _(You can also link to an open issue here)_

Current `createStructuredSelector` type definition doesn't provide auto-inferring and need users to explicitly provide generics types.

Typing definitions tested in non-strict mode

See: #428

### What is the new behaviour?

Provide a different typing strategy as suggested by @oatkiller.

Updates type definitions tests to run in strict mode

### Does this PR introduce a breaking change? _(What changes might users need to make in their application due to this PR?)_

Yes.

1. New types need Typescript > 3.1
2. New types provide working auto-inferring but **expose different custom generics definitions**. I'd expect users who provided custom generics to their `createStructuredSelector` calls to have to remove them in order to compile again with the TS compiler.

Unfortunately I wasn't able to to express the new types in a non-breaking way. It may be that is possible.

### Other information:

Parallelly working on a [similar solution for re-reselect](https://github.com/toomuchdesign/re-reselect/pull/125).

In case this gets merged in future I'd suggest the maintainers to rebuild `package-lock.json` in commit b9ea429 on their own machines.

### Please check if the PR fulfills these requirements:

- [X] Tests for the changes have been added
- [ ] Docs have been added / updated


eXamadeus:
This is a useful one to look at for #485 and #484

PR #461, opened 2020-07-19
chore(repo): update all deps (except typescript)

peter-mouland:


markerikson:
Unfortunately this has gotten too stale, especially with all the other changes I've been making. Would appreciate an updated version if you have time!

PR #465, opened 2020-07-30
Use Callable Type Signature for Memoize in createSelectorCreator

micahbales:
Addresses issue #20007 by adding a callable type signature for the memoize callback in createSelectorCreator. This avoids a type error when using Lodash Memoize as the first argument in createSelectorCreator


AprilArcus:
That issue number is from https://github.com/Microsoft/TypeScript/issues/20007

marchaos:
Any chance we can revisit this PR ?

spassvogel:
It's been a year since this repo has had any updates. Probably it's a dead end..

nickmccurdy:
I think if we were to update these types, it would be best to map them to the returned types as well.

eXamadeus:
> I think if we were to update these types, it would be best to map them to the returned types as well.

So I pulled this change into #486. I'm not sure what you mean by mapping them to the returned types. Isn't `F` the type of the input and the return? Or am I grossly misunderstanding you?

Any updates to this PR can easily be pulled into #486 or we can do the changes there. Not sure of your preference.

markerikson:
Closing since this is now included in #486 .

Issue #420, opened 2019-08-10
As of TypeScript 3.1.1, types erroneously fall back to arity-agnostic createSelector signature when used with arrays


nickmccurdy:
As far as I'm aware, this breaks any usage of Reselect's array signature with TypeScript 3.1.1+. I'm not sure what causes this yet, but I'm hoping that #366 will also fix this by using a mapped generic rest tuple. Older TypeScript versions work fine, but now that 3.1.1 is already 10 months old it seems like this may be a common issue.

Discussion: https://twitter.com/jessidhia/status/1159660672000487424
Reproduction with failing type tests: https://travis-ci.com/nickmccurdy/reselect/jobs/224127648#L1087-L1115

Issue #401, opened 2019-03-20
Version 5.0.0 proposal


josepot:
As much as I love Reselect, I hold the crazy idea that Reselect creates a leaky abstraction by exposing its 2 lowest level API functions (`createSelectorCreator` and `defaultMemoize`).

I think that if it didn't do that, then it would be possible to solve the problem of "[shared-selectors](https://github.com/reduxjs/reselect#sharing-selectors-with-props-across-multiple-component-instances)" in a way that it would be almost transparent for the developer. I discuss that idea at length in [this post](https://medium.com/@josepot/we-need-better-redux-selectors-457a1ed10732). Also, I have created [this library](https://github.com/josepot/redux-views) just to make sure that what I'm about to suggest is not totally crazy.

The API that I would like to suggest for  version 5 would be something like this:
- `createSelector`: same signature as today
- `createStructuredSelector`: like the current one, but removing its second parameter
- `createKeySelector`: an enhancer that lets the library know that the selector that is being enhanced returns an special type of result: the identifier of the instance(s) that consume the selector.

With that API it would be possible to solve the problem of shared selectors, just like [redux-views](https://github.com/josepot/redux-views) does.

Just in case that it is not obvious: I have not created Redux-Views to "compete" with Reselect. I did it just to show that it is possible to implement the API that I'm suggesting. If Reselect decides to go down this road (or a similar one) I would love to help and I would immediately deprecate Redux-Views, of course.

theKashey:
👍from a low-level API perspective, your idea is perfect. The only way I don't quite get - `.use` nobody is calling(except `.use` itself) and how you are going to clean memory.

I've also spiked a few ways to make reselect more _stable_:
- [kashe](https://github.com/theKashey/kashe) - a WeakMap based memoization. It just stores result inside the argument itself.
It's possible to have a unique pair `state` -> `selector`. For example next, I would have a unique user record from a first selector.
```js
const getUserSummary = createSelector(
  getUser,
  ....
  doSomethingWithItAll
)
// ---- something like
WeakMaps[getUserSummary].set(getUser(state, id), result); //
```
This approach hovewer does not play well with simple selectors like
```js
const getUsersInGroup = (users, group) => users.filter(user => user.group === group)
```
as long as you cannot use group(string) as a key for a WeakMap, and still able to hold only one result. But it might be not a problem.

[...]

PR #480, opened 2021-01-30
refactor(typescript): upgrade to TS 4.1, simplify and generalize types

eXamadeus:
# What

Upgrades to the latest version of TypeScript: 4.1. I _think_ this will be breaking changes. I mean it isn't a change to the code, but the TS updates might require some slight modifications to user code if they were using the `ParametricSelector` or `Selector` types directly.

## Changes

- types no longer suffer "death from a thousand overloads"
- createSelector can accept any number of selectors
	- the limit is only in the depth of recursion TS allows, which is 50 levels as of now
	- why this arbitrary number? I dunno ask the TS guys
- removed `ParametricSelector` as the base `Selector` type is now smart enough to "encompass" this role

## CI/Build Chores

- updated Travis targets to still maintained versions of node
- fixed broken build scripts
- updated TypeScript testing strategy to use https://github.com/microsoft/dtslint
	- created `/types` folder and moved tests there
	- added symlink to test folder (easiest way to setup new test)	

### EDIT: See https://github.com/reduxjs/reselect/pull/480#issuecomment-775176887 for "status" tracker of the changes suggested by everyone


.travis.yml
```diff
@@ -1,10 +1,9 @@
...
 node_js:
-      - "0.12"
-      - "4"
-      - "5"
-      - "6"
-      - "7"
+      - "10"
+      - "12"
+      - "14"
+      - "15"
```
eXamadeus, line 6: The node versions here were ancient. I updated them because the new test tooling just simply won't run in old unsupported versions.

package.json
```diff
@@ -17,7 +17,7 @@
   },
   "scripts": {
     "compile:commonjs": "better-npm-run compile:commonjs",
-    "compile:umdmin": "uglifyjs dist/reselect.js -mt -o dist/reselect.min.js",
+    "compile:umdmin": "uglifyjs dist/reselect.js -m -o dist/reselect.min.js",
```
eXamadeus, line 20: The `-t` flag is not valid, so I removed it.

package.json
```diff
@@ -42,7 +42,7 @@
       }
     },
     "test:typescript": {
-      "command": "typings-tester --dir typescript_test"
+      "command": "dtslint --expectOnly types"
```
eXamadeus, line 45: This is a superior testing suite to `typings-tester`. This is also the officially supported type testing module by the TypeScript team.

types/index.d.ts
```diff
@@ -0,0 +1 @@
+../src/index.d.ts
```
eXamadeus, line 1: This is a symlink...unfortunately `dtslint` is pretty annoying with it's lack of configurability.

In order to get the tests to run, you need the type definitions and the "tests" to be located in the same directory. Since I didn't want to touch the main `/src` directory, since it's part of the bundling path, I decided to just make a simple symlink here to let the tests run against the main type definitions.

It all works, but I would understand if there was pushback here. Although I do think that git-based symlinks work on every OS nowadays.

types/test.ts
```diff
@@ -74,62 +74,62 @@ function testSelectorAsCombiner() {
...
 
   const s: string = selector({bar: {foo: ''}});
 }
 
 type Component<P> = (props: P) => any;
 
-declare function connect<S, P, R>(selector: ParametricSelector<S, P, R>):
-  (component: Component<P & R>) => Component<P>;
+declare function connect<S, R, P>(selector: Selector<S, R, P[]>):
+  (component: Component<R & S & ([P] extends [never] ? {} : P)>) => Component<P>;
```
eXamadeus, line 89: The `Selector` types were modified, so we needed to update this helper

types/test.ts
```diff
@@ -74,62 +74,62 @@ function testSelectorAsCombiner() {
...
 
 function testConnect() {
   connect(
     createSelector(
       (state: {foo: string}) => state.foo,
-      foo => ({foo}),
+      foo => ({changeUp: foo}),
```
eXamadeus, line 95: This made the test more obvious that it was "changing" the return type values to something different.

types/test.ts
```diff
@@ -74,62 +74,62 @@ function testSelectorAsCombiner() {
...
-    const foo: string = props.foo;
+    const foo: string = props.changeUp;
   });
 
   const connected = connect(
     createSelector(
       (state: {foo: string}) => state.foo,
-      (state: never, props: {bar: number}) => props.bar,
-      (foo, bar) => ({foo, baz: bar}),
+      (state: {foo: string}, props: {bar: number}) => props.bar,
```
eXamadeus, line 107: `never` is now a proper bottom type and will not assign across a heterogenous collection of selectors.

The state object _should_ always be the same regardless of the selector, so TypeScript is doing us a favor here by forcing users to use a heterogenous type or `any`.

types/test.ts
```diff
@@ -74,62 +74,62 @@ function testSelectorAsCombiner() {
...
+      (foo, bar) => ({foo, baz: !!bar}),
     )
   )(props => {
     const foo: string = props.foo;
     const bar: number = props.bar;
-    const baz: number = props.baz;
-    // typings:expect-error
+    const baz: boolean = props.baz;
```
eXamadeus, line 113: Made the conversion of types clearer here for future developers.

types/test.ts
```diff
@@ -189,29 +189,32 @@ function testParametricSelector() {
...
-    (state) => state.foo,
-    (state) => state.foo,
-    (state) => state.foo,
-    (state) => state.foo,
-    (state) => state.foo,
+    (state: any) => state.foo,
+    (state: any) => state.foo,
+    (state: any) => state.foo,
+    (state: any) => state.foo,
+    (state: any) => state.foo,
```
eXamadeus, line 217: Non-explicit `any` is generally disallowed, we will enforce this via our type definitions. This can be disabled with a `tsconfig.json` change.

types/test.ts
```diff
@@ -225,20 +228,20 @@ function testArrayArgument() {
   const selector = createSelector([
     (state: {foo: string}) => state.foo,
     (state: {foo: string}) => state.foo,
-    (state: never, props: {bar: number}) => props.bar,
+    (state, props: {bar: number}) => props.bar,
```
eXamadeus, line 231: Implicit `any` values can be used in place of `never` for heterogenous `State` selectors.

types/test.ts
```diff
@@ -387,7 +390,7 @@ function testDefaultMemoize() {
       if (index === 0)
         return a === b;
 
-      return a.toString() === b.toString();
+      return `${a}` === `${b}`;
```
eXamadeus, line 393: TypeScript bug that was fixed...

markerikson:
Wow, thanks for putting this together!

The biggest issue atm is that @ellbee , the primary maintainer for Reselect, has been mostly inactive for a while.  Tim Dorr and I have commit rights to the repo as owners of the `reduxjs` GH org, but we don't have publish rights on NPM.  So, even if we merge this, we don't currently have a way to publish a new build.

I'm going to try reaching out to @ellbee and seeing if we can get added as publishers.

src/index.d.ts
```diff
@@ -1,966 +1,54 @@
...
-  resetRecomputations: () => number;
-}
+// Minimum TypeScript Version: 4.2
 
-export type ParametricSelector<S, P, R> = (state: S, props: P, ...args: any[]) => R;
+export as namespace Reselect;
 
-export type OutputParametricSelector<S, P, R, C> = ParametricSelector<S, P, R> & {
-  resultFunc: C;
+export type Selector<S = any, R = unknown, P extends never | (readonly any[]) = any[]> = [P] extends [never] ? (state: S) => R : (state: S, ...params: P) => R;
```
phryneas, line 5: My suggestion would be to keep `Selector`, `ParametricSelector` and `OutputParametricSelector` in as it they are, mark them as `@deprecated` and use a new name for this type to avoid a breaking change.

src/index.d.ts
```diff
@@ -1,966 +1,54 @@
...
+export type OutputSelector<S extends SelectorArray, Result, Params extends readonly any[], Combiner> = Selector<GetStateFromSelectors<S>, Result, Params> & {
```
phryneas, line 6: In the idea of backwards compatibility, make have `Combiner` stay the third argument and move `Params` to the fourth?

src/index.d.ts
```diff
@@ -1,966 +1,54 @@
...
+type SelectorResultArray<Selectors extends SelectorArray, Rest extends SelectorArray = Selectors> =
+  Rest extends [infer S, ...infer Remaining]
+    ? S extends Selector
+      ? Remaining extends SelectorArray
+        ? [ReturnType<S>, ...SelectorResultArray<Selectors, Remaining>]
+        : [ReturnType<S>]
+    : []
+  : Rest extends ((...args: any) => infer S)[] ? S[] : [];
+
+export function createSelector<Selectors extends SelectorArray, Result>(
```
phryneas, line 44: The breaking change in the arguments to `createSelector` would be kinda expected for me - I'd be okay with that.

src/index.d.ts
```diff
@@ -1,966 +1,54 @@
-export as namespace Reselect;
-
-export type Selector<S, R> = (state: S) => R;
-
-export type OutputSelector<S, R, C> = Selector<S, R> & {
-  resultFunc: C;
-  recomputations: () => number;
-  resetRecomputations: () => number;
-}
+// Minimum TypeScript Version: 4.2
```
phryneas, line 1: this should be 4.1, right?

phryneas (commented):
I added some comments. I don't think this needs a new type testing framework honestly, if anything the testing framework could be removed in the long run if we were to target only 3.9 and above.
Right now, redux toolkit, which depends on reselect goes down to 3.5, so that should probably stay supported for now though.

As for backwards compatibility: I'd suggest keeping existing types with their names as they are and deprecating them.
In addition to that, I'd keep the old type definition in a second file and use `typesVersions` to have TS choose the new one only for TS >=4.1

types/test.ts
```diff
@@ -20,10 +20,10 @@ function testSelector() {
 
   const foo: string = selector({foo: 'bar'});
 
-  // typings:expect-error
+  // $ExpectError
```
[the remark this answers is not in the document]
eXamadeus: Yeah that makes sense. I would prefer to use the `// @ts-expect-error` directive. If we need to support 3.5 for redux-toolkit, we probably need to keep the `// $ExpectError` directive though.

The main reason that I upgraded from `typings-tester` was for compatibility reasons with later versions of TypeScript _and_ for somewhat selfish development reasons. The error messages from `dtslint` were very helpful in making sure I covered all the edge cases correctly in the new typings. Seeing a descriptive error in the output was a lot easier than having to just see "error @ line ###".

I don't mind downgrading now, if you think it's important. But if we do want to support things < 3.9 we'll need some sort of tooling to help with that.

eXamadeus:
> As for backwards compatibility: I'd suggest keeping existing types with their names as they are and deprecating them.
In addition to that, I'd keep the old type definition in a second file and use typesVersions to have TS choose the new one only for TS >=4.1

I'm embarrassed to admit that I had no idea that could be done. I've never written package typings before so that thought never crossed my mind. I will absolutely do this though!

src/index.d.ts
eXamadeus: Yeah, there was a weird bug with `dtslint` that was only throwing an error in 4.1 when it shouldn't have. It literally said "change it to `4.2` to fix this problem" so I bumped it up. I am not an expert with `dtslint` thought, so there may be a more correct way to have solved that.

It did feel weird, but it was the only error when running against 4.1 and went away on 4.2 so I left it. Now that I think about this though, what would be better would be to do your earlier suggestion of splitting the types.

For the versions >=3.5,<4.0 we can use the `dtslint` set with a minimum version of 3.5 and then we can use a higher version for the "new typings".

types/test.ts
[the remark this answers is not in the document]
eXamadeus: One other note, I think the main reason that something like `dtslint` is useful is that it tests for type compatibility across a range of TS versions where the `// @ts-expect-error` only works for the installed version, right?

I think that would be a compelling reason to actually keep something like `dtslint` around, right?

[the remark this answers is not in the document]
phryneas: Over on redux toolkit, we use a test matrix that just runs `tsc` on the files with each tested TypeScript version.
https://github.com/reduxjs/redux-toolkit/blob/next/.github/workflows/tests.yml

The nice thing about `// @ts-expect-error` is the usage in the editor. If there is an error where one is expected, everything is fine. If the error is absent, you get squiggly lines. So you can just look into the editor to spot where your test breaks, you don't even need to run the tests.

Especially for writing the tests, that's a big thing.

eXamadeus:
Just as an update, I plan to get around to updating this with the suggestions this weekend. I'm sorry I didn't reply/update this sooner, it's been a busy week at work.

I am very thankful for the reviews though. I also plan to answer those and make corrections in my code as soon as I am able.

types/test.ts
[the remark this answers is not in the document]
eXamadeus: Oh, that's a good idea! I think we can do something similar with TravisCI here right?

If I have the go-ahead to do that, I don't think there are any differences between 3.5 and 3.9 (in regards to this library) so we _could_ just say that this only supports 3.5 and above while only testing 3.9 + using the `// @ts-expect-error` comments.

Basically, we can validate that 3.5, 3.6, 3.7, and 3.8 work with these typings (we know they do because of all the tests we've been running in this PR) but only setup automation to check 3.9 and above. I guess we could say this is "verified again 3.5-3.8, and actively tested against 3.9+".

I'm super new to OSS library maintenance, so I'm down for whatever you guys suggest. You're the experts haha.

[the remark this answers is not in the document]
eXamadeus: Or I can still do the matrix and use the `// $ExpectError` directive for 3.5+ and `// @ts-expect-error` for the new typings. Either way seems good to me. But the removal of the testing library seems like a more "future-proof" solution.

src/index.d.ts
eXamadeus: To bring the other conversations back here: we're going to use some form of "separated types" using `typeVersions` here, so this will no longer be necessary. We will just use the built-in `// @ts-expect-error` that was suggested instead.

src/index.d.ts
eXamadeus: Yes! Good catch. I'm not even sure why I made that change.

types/test.ts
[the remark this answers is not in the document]
phryneas: A build matrix would be possible with Travis, but since Travis CI is pretty much kicking out open source projects by restricting build times & becoming more & more unreliable, it would probably be a good idea to switch over to GH actions anyways.

As for testing older TS versions than 3.9... I'd be fine not to, since I don't expect any bigger changes coming up in the near future and the current state has already been tested for a long time. @markerikson what do you think about this?

src/index.d.ts
eXamadeus: Hmm, so I think if I address the suggestion [here](https://github.com/reduxjs/reselect/pull/480/files/66fcf9000bc173ca1f168656d2611e8448ecc188#r567287907) these _will_ be backward compatible right?

This is because the new `Selector` is just a combined version of the old `Selector` and `Parametric` Selector. The only thing that could be considered breaking, in my opinion, is the "extra args" bit of the `ParametricSelectors`. Those were always just passed through as `...any[]` where they are not fully typed. In every other instance, I think these types will behave exactly the same, no?

See this [playground link](https://www.typescriptlang.org/play?#code/PTAECpK6dqCgQXKA8gGwCagCoE8AHAUwGdQBlIgYwBcBLAewDtEw52P54iAPAhgE41QNQkTRZK6ajUEAecgBpQAJQB8oALygAFCRoBDGkQBcFAJRaNKgNxde-ISLETMqAK40CnqTPlLVZQBhDW1fWn9ldVAAMlAAb3hQUAFSd3QaADF3JiozILtk1KoGAFtvQ3pmEjMdS00NJndSgCMiAUKU0iIaFWoyiqNGJhrdesbmto74AF97PkFhUWJQAAUDAQNSnoE6KnDZAQVlVajQ3X0jUwplAgEGAlHT0AA6N42Ac1GDJjwAbQAuuNVHZuAsnMtxB4vJ51pttjRdvsiNIIkcAs8VMFznCtjs9gdImszrEEkkuiR0lkcnlQAVycUBp4htVasCmq12p1UiQen0SuVmVURmyrKAOVM7HN4JhqOgNuIAGY04WgBhYIKpK6E9FRACMUQATFEAMzKHBqHTk3mow56swYTA646qPVqRTWlF+ASGh2SL1ol0qQ3uz22wQmv1OgOHIMm0PJAUtOhMdq1Hn213KHm+1TGimR1Qm4E4D3mKPQirOgKl3QZswqA0U3PB7OkQsqYtii12WVUeWpUDK3Kq9WYTVEbUxomYputovmy1h72Z3EIpHVk76hOgG3e3Nr-HI8O64l5nd7tGFw+IgnT0+Y+MexNlZOpgTp0iZxttkgt-M8h2XYNLgZYOp4FQ3hu94us8tY6PWWbNg2AHtg2wEaD2XBIBwuEwKwyCgAAckQADuuBiGQlC0MMBF4fR4DzI4SwuNWWigD8eBROxOQANZMAwpFMCcoC8MYTCYGQqYAG7tKAAA+qiTpgzDoHgACCAibHgcicecnGAucfyrACok8OJkmgH8MntKZAD8FyVNc5DAiooC1JcxhmAEbwvAQGxbE8rl2A4izOCslY+DB5BmRZVH3pp2lRGkGQnAFpRkGJRASWQWoqUwakcb8gLBK+KbtOczoAOI9OQTmZPcpQ6iQCjukplKpWs6UkBocSJEUKXUrk+Rle+3L9IKlTDKMdRihKXIMt0vQTYMwozeykwLTMoKQhQCVaQYeDsX0Bj5WpiWHQo95qFwu01TQdVXA1ZTVpVsXZZZzopoqcmPcYGgOX94hmDZ0x3bV9WNc1CjvTle0nhdeChOSMVZXDfzfXJQTuFp2U0MovmYwIaA0AAFrZ5LJA52O40wwho5996U8koAOagZNyQz8UIwdeDMyzrOgPdQPPU1ME06kdMaIpwuQy994tez5MCDdAsC2YstPVD4s45LNCq2rINELJAjM95sOWToROgAAotIpTmICgua8YovOnbRClBoRsmztLj3WuJBu9Fb1c-D3q6b8yicco1urADawW2QTsOaD7mJz7C2hRC-s9IHwcni1ASZAwOTYNooOh+ZH3c96iPkoDSe6NbJCOwCzMOQH3UFxHPVmxQTcY0wP3ExLeME281t9Po7dq87efd9rJ5yGPUuD7Pc-zzQ+dL96LXT-joAl2XBtzxrC-wkHu+Bqv+v8+n5DkmYx8SX7Kw6tPVKI1dhdNzqiPJX0H-fa2l2LNWRskA+g9rY+UnsPOSfRSgGBTCmD4G8qYDzDjqe+DlEHIKYKg4BPNtL3wwX8PoNAcZMHwMQVqE8Xgf0Gt-ZqyUkEoKYB8NQ6DDZWQoVQmhRBWrcLMICJ+7V6bVzhjoHQvlPjfF+MCGBbdBbkCdiIgEoIZRygVEOFUwxQBUC1MYasCsiF115oAqkS5kiE2MBlERvkWG6FkQIL43l7yfwyMwhWahXKDXbuWNAEEorLycZ4w+XdL49zREXHxygZHvFcaMRhHUaDeMLr4sU4S1Cgj7AOJUejmAGKMUQExv8sEgMOpYjI1jdwKwcW8ZqAJnwGNGmmZxiS3HhzROE9Je9MkgXCWBIJMIHrRQVtUiJF9ArRMOLEjJ8SXFdJSV-XmP9+l+NSTk7CbAGK4ToigHApBhDUWFAcvZ7AuApmMAIRUBgqDiCBmSZIBgCAEGkGYfQuwOHkhaD8f5ZgWgMHVJOFgiYNj3BoCDTaptpQlBGMIS8aS3nSHYjoTARgDDeScsCDFhgXivPeUQeA8KgFIoAEL-KYAYNFeKsUUBxWKOlLw-nUupSS6oiKYxBAhQwYQ2h0WYuxVcW49xHhmHiKAFMFQzB8QEkJUAMxlDABQNS7YnzbwcIgMAXFmKXhUF5TQLgpLhB3GuXVb5Hw0XSQMOgdw1wvmoOBKSkFLx0AMA+DoG1dqiDmA5Qi0AZq6bkuBdIH41rbX2sBaG0FzrqiuvdZ6719q-UmsDd8mgREYURp9dCzkAg40jATR6r1kbfXGs5eKMiOp2KGMnMY+8VpkhIvUiiogLSKVUoMB27lhqWk6EJdIZQrL-nKANVpPlwJ+osyDQ9TVnrB3loFrOkNIKfg6BHdSv1y6M1ZvzTocdkLLDkjmH6-1QCxw1u0GOCcU4TxNrqbaVtRKe22kpWy7ty5aA8onfjckA623Dq7WOw1U7mazotaggDRLt0zozausNTAN1dtg8kWde6pgHtA6AE98Az1IBwKTOgZBiNFVEhO4m8rQBtANe4Xk4pmDiAYIqEQ5NH09EEGQe5lDbWFUMLxcQNLk1EF8h8BgoAaRlG2HTVj4h5RAP8vCNUTBQDoHKqAAAnBpoq2AyakdInQdA6BQCkUELxeAqZSI6h0PERdZgADk5M8D2aAx+swiJ7Ugd-WYE0SqEhSqYDK0A9m8DqgYPZxVyh7MtHlKTezfqkApiHNx2OwgDNGdabjGgfGXAc0DelHockDBAtksoHIamBOybVFgUAJsSD6LJkYUAAArOjwhbWkUOmQUmBgyAAANZHFQBH1jiZA8uKhTLajirjmh43CsSy9jbbNtoc05lz1Gu3uYEJ5gxhqfN+cldKzwDnQvuoi356LsX4sEQAJL2dKBJkYONxB0BY3pkjXHQAACIWhal4oQqgPWOFEC+2qYmr2pU0Hs2QVrQCaUtHcFaxrwhSO0Y+KTGgLxQA3ZY8QB4qKdF3AYNJOgmBCFIs46xpr-HxBiU2AFioJBlDvZM4Z4zsOjVICJyT2U1Geg3Pm7uAwP1RBDkELJ0ovkgA) for an example of what I'm talking about.

src/index.d.ts
eXamadeus: I'll have this fixed in the next push.

types/test.ts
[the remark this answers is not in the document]
eXamadeus: If @markerikson is good with it, I'll do the switch over to GH actions here too. It's the way of the future anyway, right? 😆 

I could also totally break that up into another PR. Just let me know what you would prefer.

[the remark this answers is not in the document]
eXamadeus: > As for testing older TS versions than 3.9... I'd be fine not to, since I don't expect any bigger changes coming up in the near future and the current state has already been tested for a long time.

That would also make the testing a lot easier because you wouldn't need to do the swapping of the test comments or maintain separate type files. I was looking at the way you guys did it in the toolkit and that last bit of the job looked a little gnarly with all the `sed` stuff.

markerikson:
Hey, wanted to follow up on this.

I'd like to split the work in this PR into a couple different PRs if at all possible.

First, per #482 , I definitely want to modernize the CI setup, including switching from TravisCI to Github Actions and dropping old versions of Node.  However, I think that should be its own PR.  

From there, yes, I do want to update any dependencies and actual TS typings.

Not sure if it'd be easier to close this PR and then create two new ones?

markerikson:
Also: my other concern atm is that it looks like some of the changes in here are probably enough to justify a Reselect 5.0 release.  However, if we do a 5.0, it seems like there are also a number of other functionality changes that should be considered as well, such as #401 , and even rewriting the code itself in TS.

Meanwhile, there's open PRs like #465 , which seems small but useful, and could be put out in a new 4.x release.

Given that, can we figure out a way to scope the changes so that we can have a 4.x release soon with some additional bits of cleanup, and then do a larger effort to work towards a 5.x?

eXamadeus:
> Given that, can we figure out a way to scope the changes so that we can have a 4.x release soon with some additional bits of cleanup, and then do a larger effort to work towards a 5.x?

Sure. I really like that idea. We can do a 4.x PR that is only non-breaking upgrades. A non-exhaustive list:
- CI upgrade from Travis -> GHA
- #465
- ...whatever else looks good to pull into a 4.x release.

Then I can start a 5.0 release w/:
- the 4.1 type updates I did here
- some incarnation of #401?

I would love to include #401 (well some variant of it) in the 5.0 release PR, but I'm not sure there is a consensus on the names for the new API. I have read through all the posts and related issues (that took a while BTW 😆) and think it's a great idea. I'm of the opinion it shouldn't be called `createKeySelector` (which I think @josepot agrees with). It should still behave the same as the idea you proposed and will allow for things like shared selectors. Heck, it might even enable things with can't think of right now. I really like that idea!

My personal opinion is that it should be called `createDynamicSelector` **AND** that it should still be able to accept the `state`. The "dynamic selector" moniker would just let people know that it won't be considered for caching purposes. This is a slightly less "leaky" way to name it. What do you think @josepot?

I have no problem making both of these PRs. Or I can contribute to them if you'd rather control that process. Just let me know.

markerikson:
@eXamadeus yep, sounds good. Let's do the build updates first. In fact, let's split this up into separate PRs for :

1. CI changes
2. Dependency updates
3. Any minor fixes we feel are appropriate for a 4.x release.

I think it's worth having a larger planning discussion around what 5.x should include. There's a bunch of open PRs that I could see conflicting. I think it's also important to analyze what weaknesses Reselect currently has and what use cases it doesn't convert well atm, and look at other alternatives in the ecosystem for comparison. 

Related to all this, I've been looking at https://github.com/dai-shi/proxy-memoize as a possible better alternative to Reselect that we could start recommending.  Discussion thread for that topic is at https://github.com/reduxjs/react-redux/issues/1653. 

However, I'm also interested in improving Reselect itself too. If we're going to do so, I want to have a plan laid out. 

I'll try to create a discussion thread tonight or tomorrow.

eXamadeus:
@markerikson Sounds good, I'll have them up tonight or tomorrow (at least the CI one for sure).

Question about the dependency updates: which updates does `reselect` want to entertain? If we decided to just assume TS 3.9+ support we can drop the testing libraries completely. If that's the route we want to go, that's probably the easiest one. Also, do you want to do a generic upgrade of other dependencies while were at it? Or just remove the testing libraries and update to say 3.9+ is supported?

Either way, I can create the CI PR and a "minor fixes" PR that follows that w/ #465 in it.

markerikson:
Yeah, I still want to do this in phases.  There were several open "update deps" PRs I closed today.  Like, this is apparently still on Babel 6 (!).  

I think we can certainly ignore TS < 3.5, and it's probably not unreasonable to go 3.9+.  Of course, as we just saw, 3.1+ appears to be an issue :)

eXamadeus:
> I think we can certainly ignore TS < 3.5, and it's probably not unreasonable to go 3.9+. Of course, as we just saw, 3.1+ appears to be an issue :)

Yeah, it's only two of the tests that are broken (sort of). The actual library in production is already affected by this, so it's not like removing the tests would cause a breaking change. It's basically just acknowledging that the types work as expected for 2.9 and 3.0, but don't behave correctly for 3.1+. Maybe we should make a bug fix for that in 4.x?

The new typings for TS 4.1+ do actually work for this tested scenario, so the test is brought back for TS 4.1+. If you're alright with it, we can just make this PR only do the type upgrade to TS 3.9+ with no breaking changes (so we can keep this as reselect v4)? I am still planning to make that other PR that will be for the 5.0 release w/ TS 4.1+ typings.

Of course, both of these would require the work done in #483 to use the new CI changes. Let me know if you want me to repurpose this PR or create another one for the "reselect v4 type updates (TS 3.9+)".

markerikson:
Hah, I'm kinda getting lost on what's where now :)

Yeah, let's try things in this order:

- Turn off the couple failing tests in the GH Actions PR, as the failing tests are not due to the GH Actions switch, and get that merged in
- Update any other dependencies and build tooling
- Figure out if there's a reasonable fix to the types that could be made in a 4.x release
- Hold off on a larger rewrite of the types, because this will be affected by any noticeable changes to the API that might be involved in a v5

markerikson:
@eXamadeus: so, it turns out that this bug was already reported by @nickmccurdy two years ago :)

https://github.com/reduxjs/reselect/issues/420

types/test.ts
nickmccurdy: I noticed that `R` and `P` are swapped here. Would that impact back compatability?

types/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
+    "noEmit": true,
+
+    "baseUrl": ".",
+    "paths": { "src": ["."] }
+  }
+}
```
nickmccurdy, line 14: Can we use `"strict": true` or would it be too noisy?

nickmccurdy (changes requested):

nickmccurdy:
Whoops sorry, I reviewed this without reading the discussion first. I'm now following the repo so if there are new PRs I should be able to review them. Let me know if any other help is needed.

types/tsconfig.json
eXamadeus: Yeah, I think that would be fine. This was the recommended `tsconfig` for the old testing library (typings-tester?) I didn't think about updating it, but using strict is definitely worth it IMO.

I'll make sure that's added to whatever final PR goes in. Thanks for the catch!

types/test.ts
eXamadeus: Not here in the test, but it might cause issues in client usage of `Selector`. This was something that I should revert though.

I changed it without realizing because I rewrote the types from scratch. I'm definitely going to fix the order here, since it's a pointless change that would just inconvenience people.

eXamadeus:
Hey @nickmccurdy this is still the best one to review the type changes I suppose. But @markerikson brings up a good point that these type changes need to wait until a v5 release.

Here's a list of things I'm planning on doing now:

- [ ] Turn off the couple failing tests in the GH Actions PR (#483), as the failing tests are not due to the GH Actions switch, and get that merged in
- [ ] Retarget (recreate?) this PR to `v5.0.0` branch. @markerikson I think these types will be best to get in _before_ the API changes, since they should be:
    1. easier to modify
    1. shouldn't conflict with the changes suggested in #401
    1. might actually help the changes in #401 be easier to implement
- [ ] Figure out any type changes that can be in 4.x and make a PR for those. Ones that seem likely:
    1. #465
    2. Seeing if I can find a fix for #420 
    
For the deps PRs, I can do those too. Looks like #461 and #381 look like good candidates there. I would think the babel upgrade (#381) should probably go first, then #461 can go on top of that. Would you like me to pull those together in a deps PR as well? Or is that something you guys want to do?

markerikson:
@eXamadeus Yep, please go ahead and do a deps PR as well if you have time.

I _really_ appreciate your effort and contributions here!

eXamadeus:
I'm going to make issues to track the "bullet points" I made earlier and I'll edit the comment to link to them.

markerikson:
FYI, we do already have an "upgrade Babel" PR at #381 , which Andarist just rebased.  However, it now conflicts with the removal of the Travis file. 

Not sure what the best option is for resolving that atm.

eXamadeus:
> FYI, we do already have an "upgrade Babel" PR at #381 , which Andarist just rebased. However, it now conflicts with the removal of the Travis file.
>
> Not sure what the best option is for resolving that atm.

I'll comment on that PR, but he just made the same "changes" we just made to the node versions essentially. It looks like just deleting the `.travis.yml` file in his PR will resolve the conflict and should "just work"&trade;.

markerikson:
Good call! Just did that via the web UI and I see the GH actions kicking in now.

eXamadeus:
Whoops, I should have made a branch on my fork. When I updated my `master` branch it purged this PR of the code.

The latest commit here for reference is 66fcf9000bc173ca1f168656d2611e8448ecc188. You can see the changes, as they originally were, @ https://github.com/reduxjs/reselect/pull/487.

PR #483, opened 2021-02-08
ci: replace TravisCI with GitHub Actions

### LINKED ISSUE #482: Modernize CI setup
TravisCI appears to be mostly dead for OSS projects at this point.

We should switch to Github Actions instead.  Also interested in any other updates that would be beneficial, like dropping support for outdated Node versions and testing the TS typings against multiple versions of TS.

It may also be worth borrowing some setup from Redux Toolkit at this point:

https://github.com/reduxjs/redux-toolkit/blob/ba8c911c53e67365499ac00982f640b369360a64/.github/workflows/tests.yml

eXamadeus:
## Replace TravisCI for GitHub Actions

This pulls inspiration from https://github.com/reduxjs/redux-toolkit/blob/next/.github/workflows/tests.yml and spawned from conversations in #480.

### Fixes #482

Notes to check up on:
- [x] ~are the number of TS versions correct~ **this is interesting, see the Notes section**
- [x] should we remove some of the older node versions?
- [x] will this actually work once we merge into the main repo? or do we need to enable actions somehow?

## Notes

I was going to update TypeScript to 3.5 (and only test past there) as we discussed in #480 but it looks like there is an interesting bug with TypeScript (or maybe a fix that seems like a bug) where the array typings used by `reselect` to type the combiner...well just don't work as expected past TS 3.1...but that makes sense. We should really do that in the type update PR anyway (the 5.0 release).

<details>
<summary>More info...</summary>

> Upon looking into this, it appears to just have been a change in the way TypeScript resolves the order of inferred type parameters. It checks the combiner function types before enforcing the constraints of the selectors. This means that the types for array-input selectors probably never worked as intended for TypeScript 3.1+.

</details>

I have the failures showing here in CI: https://github.com/eXamadeus/reselect/actions/runs/546546429


eXamadeus:
Let me know what you think about the questions I left up @markerikson and @phryneas

eXamadeus:
See passing CI here: https://github.com/eXamadeus/reselect/actions

(and all the failing ones from my testing...haha)

markerikson:
Yipes. That's... lovely :)

Obviously my goal here was that we'd do the CI switch, see the current code building okay, and then work on any further updates.  

I don't have an immediate suggestion here, tbh.  I know Lenz has been kinda burned out lately, so while I'd _really_ like to get his feedback on this, not sure if he's feeling up to it atm.

I'll try putting out the bat-signal on Twitter and see if anyone else has suggestions.

nickmccurdy:
> should we remove some of the older node versions?

Looks like this was already done. I agree with it as only Node 10+ is currently supported.

> will this actually work once we merge into the main repo? or do we need to enable actions somehow?

Yup, it should be automatic.

nickmccurdy (approved):
Looks good overall, though I'm not sure if we need both Coveralls and Codecov, they both gather code coverage data.

eXamadeus:
Yeah I just brought over everything that was in Travis. I didn't want to remove anything haha.

Also, I'm a dope. I totally changed the node versions already. What I meant to say is "are these the ones we should support" or something haha!

nickmccurdy:
I'm not sure. It may be safer to leave them both in, unless another maintainer can confirm if only one is needed.

eXamadeus:
OK so I _think_ I verified that Codecov.io is uploading correctly: https://codecov.io/github/eXamadeus/reselect/commit/9626793b028b2872bca2a02cc3353ba6c7ad10ad

Although I'm not entirely convinced, because I would have expected a comment on this PR...any ideas how to set that up? The [docs for the action](https://github.com/codecov/codecov-action#usage) don't lead me to believe there is a configuration value I'm missing.

eXamadeus:
Hmm, well it's working on my "testing" PR: https://github.com/eXamadeus/reselect/pull/1

Not sure if this is an issue with forked repos vs "natural branches" on the main repo...

eXamadeus:
@nickmccurdy Are you familiar with the comment setup? I made some small changes to make sure Codecov reports when PRs are made. Let me know if those settings make sense.

I decided to turn off the "master only" filter so branches like `v5.0.0` would get coverage checks too. Not sure if that's desirable or not.

markerikson:
Yeah, if the build+test steps are working here, I'm fine with getting this in as a baseline.  We can always double-check code coverage behavior later and fix it if necessary.

markerikson:
Skimming the settings, seems reasonable to my not-familiar-with-GH-actions eye.  You happy with this as-is?

markerikson:
Okay, let's get this in and keep going, then.

PR #486, opened 2021-02-08
TypeScript update for 4.x

### LINKED ISSUE #485: TypeScript updates for v4.x
## Upgrade Types without breaking changes

We'd like to perform any updates/fixes to TypeScript that _won't_ result in breaking changes.

#### Goals
- fix bugs w/ typings (#384 is an example)
- support TypeScript 3.9+ and validate it in CI
- remove failing tests for 3.1 changes (#420) _OR_ fix those issues in the typings

---

### Some caveats

- #483 replaces the CI pipeline with GitHub Actions. We need to support testing multiple versions of TypeScript with that new CI structure (this is already worked out in #480)

eXamadeus:
## Update TypeScript without causing breaking changes

Upgrade TypeScript to 3.9 and fix any broken issues that can be fixed without a major version update. Any PRs that can be included to fix the typings without a major version update will be pulled into this PR.

### Includes PRs:

<details>

<summary>#465: Use Callable Type Signature for Memoize Callback in createSelectorCreator</summary>

Addresses issue https://github.com/microsoft/TypeScript/issues/20007 by adding a callable type signature for the memoize callback in createSelectorCreator. This avoids a type error when using Lodash Memoize as the first argument in createSelectorCreator

</details>


markerikson:
Side note: why are we getting _three_ different code coverage report comments? :)

eXamadeus:
> Side note: why are we getting three different code coverage report comments? :)

Haha, good question! There is a baseline of two because Codecov and Coveralls are both running (lol, obviously).

I _guess_ that Coveralls is having trouble distinguishing the main repo from the fork 🤷? Regardless, seems like a good reason to drop Coveralls and use Codecov instead. WDYT?

I can make a separate PR to pick one of the two real fast. It's just a few lines in the workflow file.

eXamadeus:
Side note: HOLY SMOKES BATMAN, LOOK AT THOSE LINES REMOVED FROM THE PACKAGE-LOCK.JSON!
image

markerikson:
I'm for whatever provides the most useful info and the least amount of annoyance :)

As far as I know Reselect is the only one of the Redux org repos that has code coverage reports turned on in PRs, and given the very small size of the library, I'm not sure I find that all that useful. (Granted, I'm also not the person who spent most of their time maintaining this lib in the first place :)  )

markerikson:
Code-wise, this seems reasonable so far. Looks like you've mostly swapped the type tests for `tsc` itself, and tweaked some of the arg definitions in how they extend functions?

eXamadeus:
> Code-wise, this seems reasonable so far. Looks like you've mostly swapped the type tests for tsc itself, and tweaked some of the arg definitions in how they extend functions?

Yup! And set the baseline TS version to 3.9 (mainly to make the testing easier)

---

> I'm for whatever provides the most useful info and the least amount of annoyance :)
>
> As far as I know Reselect is the only one of the Redux org repos that has code coverage reports turned on in PRs, and given the very small size of the library, I'm not sure I find that all that useful. (Granted, I'm also not the person who spent most of their time maintaining this lib in the first place :) )

I'll make a PR to remove Coveralls, and I guess if anyone gets angry I'll just blame it on you 😆

On a more serious note: we could leave the "removal" PR up for a while and see if someone really feels adamant about it. Give it a week or so and if no one says anything we could merge it right? It's super easy to add back in whenever.

I'm also not convinced Coveralls was even running anytime recently:

image

eXamadeus:
Side note: made #488 to track discussion on the coverage reporters.

markerikson:
Just published this current PR as https://github.com/reduxjs/reselect/releases/tag/v4.1.0-alpha.0 for folks to try out.

markerikson:
Good enough. Merging this, and following up with a PR that actually converts the source to TS.

PR #499, opened 2021-06-11
feat: createStructuredSelector returns OutputSelector

stropho:
Hi, 
unlike `createSelector` that returns `OutputSelector`, the `createStructuredSelector` currently returns a `Selector`.
What is the difference?
`OutputSelector` is extended `Selector` and additionally, it includes:
```
  resultFunc: C;
  recomputations: () => number;
  resetRecomputations: () => number;
```
Those are great for testing purposes. These attributes are obviously there, but it is a bit awkward to access them in TS project.
The `resultFunc` accepts the the input selectors in somewhat unknown order, because `Object.keys` does not guarantee to return the same values in the same order in all environments. Thus the arguments of `resultFunc` in this PR are just `any`. Nonetheless e.g. `recomputations` are easily accessible.


markerikson:
Still useful, but the PR is stale vs `master`, and the latest types are... tricky to say the least.

stropho:
Thx for the note. I wouldn't notice the change on master after quite a long time :)

PR #500, opened 2021-06-28
Add TypedCreateSelector interface to support creating a pre-typed version of createSelector

ICodeMyOwnLife:
Similar to [`TypedUseSelectorHook`](https://react-redux.js.org/using-react-redux/usage-with-typescript#define-typed-hooks), this interface allows us to easily create a version of `createSelector` that is properly typed for our store's root state.

Code example:
```ts
import { createSelector, TypedCreateSelector } from 'reselect'
import type { RootState } from './store'

export const createAppSelector: TypedCreateSelector<RootState> = createSelector
```

This frees us from having to type selectors all the time in some codebase:

Instead of
```ts
const selector1 = createSelector((state: RootState) => state.users, ...)
const selector2 = createSelector((state: RootState) => state.roles, ...)
```

Now we can do this and it's still type-safe:
```ts
const selector1 = createSelector(state => state.users, ...)
const selector2 = createSelector(state => state.roles, ...)
```

There is also a question on this feature on StackOverflow: [How to define a version of createSelector typed for my app?](https://stackoverflow.com/questions/67145868/how-to-define-a-version-of-createselector-typed-for-my-app)

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -16,951 +16,1557 @@ export type OutputParametricSelector<S, P, R, C> = ParametricSelector<S, P, R> &
   resetRecomputations: () => number;
 }
 
-/* homogeneous selector parameter types */
+export const createSelector: CreateSelector;
 
-/* one selector */
-export function createSelector<S, R1, T>(
-  selector: Selector<S, R1>,
-  combiner: (res: R1) => T,
-): OutputSelector<S, T, (res: R1) => T>;
-export function createSelector<S, P, R1, T>(
-  selector: ParametricSelector<S, P, R1>,
-  combiner: (res: R1) => T,
-): OutputParametricSelector<S, P, T, (res: R1) => T>;
+interface CreateSelector {
+  /* #region Homogeneous selector parameter */
+  /* 1 selector(s) */
+  <S, R1, T>(
+    selector1: Selector<S, R1>,
+    combiner: (res1: R1) => T
+  ): OutputSelector<S, T, (res1: R1) => T>;
+  <S, P, R1, T>(
+    selector1: ParametricSelector<S, P, R1>,
+    combiner: (res1: R1) => T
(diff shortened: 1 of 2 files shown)


nickmccurdy:
Would this still be compatible with old types when not giving the state?

ICodeMyOwnLife:
> Would this still be compatible with old types when not giving the state?
From the view of library consumer this change only adds a new utility type and does not change the old types.
This change passed all old typing tests.
I also have tested it (by replacing the declaration file in `node_modules` with this new file and running type-check) against projects of my organization and have seen that it didn't introduce any regression. Those projects are enormous codebases with wide use of `createSelector`.
So I'm quite sure that this would be compatible with old types.

Before, people do this
```ts
// someSelector.ts
import { RootState } from 'somewhere';

const someSelector = createSelector((state: AppState) => someState, someCombiner);
```

After, people can still do as before, but they have one more option
```ts
// utils.ts
import { TypedCreateSelector, createSelector } from 'reselector';
import { RootState } from 'somewhere';

export const createAppSelector: TypedCreateSelector<RootState> = createSelector;

// someSelector.ts
import { createAppSelector } from 'utils.ts';

// `state` will be inferred to `RootState`
const someSelector = createAppSelector(state => someState, someCombiner);
```

gheorghesava-mck:
Please make sure that the return type is preserved even after applying some equality checks like shallow equal or deep equal. I observed that using useSelector hook with one of the equality checks loses the type of the returned value, being transformed to any

markerikson:
@gheorghesava-mck that sounds like an entirely different problem. Can you post a separate issue for that, and please include a CodeSandbox or repo that shows the problem happening?

gheorghesava-mck:
@markerikson just tested with the latest version and it worked fine. Seems I was behind with some patch versions for react-redux. My apologies!

siarheipashkevich:
Any news?

PR #514, opened 2021-10-20
Additional 4.1 tweaks based on existing PRs

### LINKED ISSUE #350: More informative error message for createSelector
I've just had a case when I misused `createSelector` by mistakenly passing `resultFunc` into array with inputSelectors.
```js
createSelector([
  firstSelector,
  secondSelector,
  (firstParam, secondParam) => {
     ...
     return { ...derivedProps }
  },
])
```
And error message I had is bit confusing   
```
resultFunc.apply is not a function
```   

and it took me a while to define what exactly caused this error.

I propose to add error in case when resultFunc is not a function to make it more clearer to find and fix mistakes like this.

### LINKED ISSUE #404: Improve error exception for debugging circular dependencies
Currently, if you pass an undefined dependency into createSelector(), reselect correctly throws an exception. The string in that exception can be made a little bit more clearer, which would be helpful for clients that are hitting this and trying to find the source of the problem.

[Here ](https://github.com/reduxjs/reselect/blob/master/src/index.js#L41)is how the error string is computed right now. An easy way to make this better would be to change this

```javascript
    const dependencyTypes = dependencies.map(
      dep => typeof dep
    ).join(', ')
```
to this

```javascript
    const dependencyTypes = dependencies.map(
      dep => typeof dep === 'function' ? `${dep.name}()` : typeof dep
    ).join(', ')
```

Circular dependencies are a common pain point in Javascript apps. If you google this error string you will find articles about people trying to solve this issue. It's not reselect's fault at all, but reselect can help make them easier to find if they happen to surface at the reselect layer.

https://decembersoft.com/posts/error-selector-creators-expect-all-input-selectors-to-be-functions/

markerikson:
This PR:

- Adds an error message if `resultFunc` isn't a function
- Improves the error message for bad dependency args
- Includes `memoizedResultFunc` and `lastResult` in the util fields of the selector

Fixes #350 
Fixes #404 

Supersedes #358 , #359, #405 , #448 .

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -62,6 +62,7 @@ export function createSelectorCreator<
   // (memoize: MemoizeFunction, ...memoizeOptions: MemoizerOptions) {
   const createSelector = (...funcs: Function[]) => {
     let recomputations = 0
+    let lastResult: unknown
 
     // Due to the intricacies of rest params, we can't do an optional arg after `...funcs`.
     // So, start by declaring the default value here.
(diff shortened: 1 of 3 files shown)


RusinovAnton:
Thank you!

PR #535, opened 2021-11-02
Update defaultMemoize cache even if resultEqualityCheck is a hit

tetslee:
I tried out `resultEqualityCheck` and it caused my selector to keep recalculating because the cache doesn't get updated.


markerikson:
Thanks, could you add a test for this behavior as well?

markerikson:
Thanks!

PR #537, opened 2021-11-02
Clean up OutputSelector typing and fix bug with memoize function types

### LINKED ISSUE #534: TypeScript error in 4.1: `Type instantiation is excessively deep and possibly infinite`
After upgrading to 4.1, we're seeing the following TS error among many of our selectors:
> Type instantiation is excessively deep and possibly infinite.

We rely heavily on selector composition in our app so some of these are deeply nested (as in, their input selectors have input selectors that have input selectors etc.) - 7-8 levels deep for most of the offending ones. I haven't been able to find specific repro steps, but it doesn't seem to happen with simple/more shallow selectors.

Are there any constraints/limitations with the new types? Are there any new best practices in how we define our selectors that can help avoid this? All `createSelector` use inferred types (we don't define the type parameters manually anywhere).

eXamadeus:
# Pass memoize function types to selectors

- Fixes selector typing issue discovered while looking into #534 
- Adds FAQ entry about deep type instantiation errors
- Cleans up types for deeply nested selectors

Selectors created with the default `createSelector` should use the `defaultMemoize` function internally as their cache. The `defaultMemoize` function attaches a `clearCache` function to each selector created. This type information was not being passed to those selectors correctly. Now it is 😄!

### BEFORE: selector can't find the `clearCache` provided from the `defaultMemoize` function
Image 2021-11-02 at 1 45 43 PM

### AFTER: the type is passed to the selector correctly
Image 2021-11-02 at 1 46 59 PM

## Cleans up type information when using deeply nested selectors

This was found when I was trying to fix #534. Basically, the type information was hideous, but while looking into the problems I found the bug above. Fixing it, fortunately also cleaned up these types.

### BEFORE: types are kinda hideous
Image 2021-11-02 at 1 45 03 PM

### AFTER: much more legible
Image 2021-11-02 at 1 47 10 PM


typescript_test/test.ts
```diff
@@ -1129,3 +1139,49 @@ function testInputSelectorWithUndefinedReturn() {
...
+  const selector19 = createSelector(selector18, s => s)
+  const selector20 = createSelector(selector19, s => s)
+  const selector21 = createSelector(selector20, s => s)
+  const selector22 = createSelector(selector21, s => s)
+  const selector23 = createSelector(selector22, s => s)
+  const selector24 = createSelector(selector23, s => s)
+  const selector25 = createSelector(selector24, s => s)
+  const selector26: Selector<
+    // this can be created as a utility type, or we could expose this from the
+    // library since it already exists...
```
eXamadeus, line 1180:
```suggestion
```

test/test_selector.ts
```diff
@@ -735,11 +735,22 @@ describe('defaultMemoize', () => {
...
+    // Already in cache
+    selector('a') // ['c', 'b', 'a']
+    expect(funcCalls).toBe(4)
+
+    // make sure clearCache is passed to the selector correctly
+    selector.clearCache()
+
     // Cache was cleared
     // Note: the outer arguments wrapper function still has 'c' in its own size-1 cache, so passing
     // 'c' here would _not_ recalculate
```
eXamadeus, line 751:
```suggestion
    // Note: the outer arguments wrapper function still has 'a' in its own size-1 cache, so passing
    // 'a' here would _not_ recalculate
```

eXamadeus:
@markerikson The comments are fantastic! Thanks for adding them and cleaning up some of the type parameter names.

typescript_test/test.ts
```diff
@@ -1164,8 +1164,8 @@ function deepNesting() {
   const selector17: OutputSelector<
     [(state: State) => string],
     ReturnType<typeof selector16>,
-    never,
-    (s: string) => string
+    (s: string) => string,
+    never
```
eXamadeus, line 1168: Good catch, thanks!

eXamadeus: I knew this would cause an issue if the other PR merged, but forgot to mention it.

markerikson:
As a side note, last night I was playing around locally with refactoring a bunch of the typedefs with two goals:

- Extract duplicated types into something reusable, like the "result of `createSelector()`" types that are all identical
- Make the final hover preview types of a selector more readable.

I didn't come up with any good concrete results. I was able to successfully extract some types, but I found out that doing so just adds another level of indirection to the hover previews.

I then tried using some different "expand/unpack types down to the final fields" utility types in various places to simplify the hover previews. Those gave me mixed results.  I was sometimes able to see the final fields more clearly, but some attempts resulted in more "excessive instantiation" errors in the typetests, others broke everything because the utility types weren't including call signatures in the output.

I was also wondering if there's some way to convert `Selector` into an `interface` so that `OutputSelector` could in turn `extends Selector<X,Y>, OutputSelectorFields<Y,Z> {}`, but I'm not sure if it's possible because of the need to switch between those two different call signatures.

As usual, I just barely know what I'm doing here and am sorta randomly fiddling with things to see what happens :)

markerikson:
Another thought:

I know that the selector itself does have a `.clearCache()` function attached by default, because we're using `defaultMemoize()` on both the "combiner" function and the "args comparison" wrapper.

What's the actual difference in behavior if you call `selector.clearCache()` vs `selector.memoizedResultFunc.clearCache()`?


(other side note: I've always referred to the `createSelector` arguments as the "input selectors" and "output selector", rather than calling it a "combiner" or a "result func", but the name `OutputSelector` is already taken in the types to refer to "the function returned from `createSelector`. This is an annoying name clash that bothers me.)

markerikson:
This is more than good enough for now, let's get this out.

markerikson:
Out in https://github.com/reduxjs/reselect/releases/tag/v4.1.2

eXamadeus:
Sorry I was slow to respond. There are definitely some other ideas along the same vein as the changes you were fiddling with. Another interesting thing is the new version of typescript might include tail-recursive optimizations, so we might be able to change the types around to avoid the "infinite recursion" issues altogether (not sure how easy it will be to make them all tail-recursive though).

markerikson:
@eXamadeus you said in this PR and the other issue that you were able to "bump the recursion limit from 10 to 15 selectors". I've spent my whole day hacking on the types in relation to another issue.  if I revert all the changes and go back to 2a314fd , I'm seeing it start throwing "excessively deep" errors on `const selector11 = createSelector(selector10, s => s)`.  I also don't see anything that looks like an actual change to recursive behavior anyway.

Did you actually make a change along those lines? If so, what am I missing?

eXamadeus:
> @eXamadeus you said in this PR and the other issue that you were able to "bump the recursion limit from 10 to 15 selectors". I've spent my whole day hacking on the types in relation to another issue. if I revert all the changes and go back to [2a314fd](https://github.com/reduxjs/reselect/commit/2a314fd2b0e179c76167090caf8465575fee955f) , I'm seeing it start throwing "excessively deep" errors on `const selector11 = createSelector(selector10, s => s)`. I also don't see anything that looks like an actual change to recursive behavior anyway.
> 
> Did you actually make a change along those lines? If so, what am I missing?

Sorry I didn't make the change anywhere. It really made the types quite ugly, so I left it out.

I think there are some changes upcoming in the newest version of TS (4.5) that will allow us to "increase" this limit by using tail-recursion (which I'm not sure how easy that will be w/ our types here). There's some more info here: https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-rc/#tailrec-conditional

I think the stuff above is really promising.

eXamadeus:
> Another thought:
> 
> I know that the selector itself does have a `.clearCache()` function attached by default, because we're using `defaultMemoize()` on both the "combiner" function and the "args comparison" wrapper.
> 
> What's the actual difference in behavior if you call `selector.clearCache()` vs `selector.memoizedResultFunc.clearCache()`?

I think they're the exact same method, I think it's just a form of "utility"?

> (other side note: I've always referred to the `createSelector` arguments as the "input selectors" and "output selector", rather than calling it a "combiner" or a "result func", but the name `OutputSelector` is already taken in the types to refer to "the function returned from `createSelector`. This is an annoying name clash that bothers me.)

Yeah, I'm not sure what to call this. Output selector does make sense, but as you said, that's the result entirely. Renaming it might be more trouble than it's worth, but FWIW it bothers me too.

markerikson:
Thing is the "selector" is the outer function that just checks the arguments:

https://github.com/reduxjs/reselect/blob/v4.1.2/src/index.ts#L115

whereas the "memoized result func" is generated separately:

https://github.com/reduxjs/reselect/blob/v4.1.2/src/index.ts#L105

PR #542, opened 2021-11-07
[DRAFT] Try rewriting state/parameter inference for correct intersections

### LINKED ISSUE #540: Multiple selectors with arguments result in a union type
Typescript 4.4.2
Reselect 4.1.1

If I have a selector like:

```
export const testSelector = createSelector(
  (_: RootState, { testNumber }: { testNumber: number }) => testNumber,
  (_: RootState, { testString }: { testString: string }) => testString,
  (testNumber, testString) => testNumber + testString
)
```

When I try to call the selector using:

```
  const test = useSelector((state: RootState) =>
    testSelector(state, { testNumber: 1, testString: '10' })
  )
```

I get the error:

```
Argument of type '[{ testNumber: number; testString: string; }]' is not assignable to parameter of type '[{ testNumber: number; }] | [{ testString: string; }]'.
  Type '[{ testNumber: number; testString: string; }]' is not assignable to type '[{ testNumber: number; }]'.
    Type '{ testNumber: number; testString: string; }' is not assignable to type '{ testNumber: number; }'.
      Object literal may only specify known properties, and 'testString' does not exist in type '{ testNumber: number; }'. ts(2345)
```

It appears that Reselect is converting all the different arguments into a union type, when instead they should be combined using an intersection type.

Workaround is to drop back to Reselect 4.0.0, which uses the intersection type.

markerikson:
### THIS IS EVEN MORE WIP AND BROKEN THAN USUAL DO NOT MERGE

This is supposed to be an attempt to fix #540 .

Between Lenz and myself, we've come up with something that actually seems to mostly work.

I actually also had a version that was using `List.MergeAll>` from `ts-toolbelt`, but it seems to be losing some tightness in the process - a whole bunch of typetests that are _supposed_ to fail were no longer failing.

At the moment, this is down to two errors:

- Something with the `Selector` type is wrong. I don't think the `Params extends [never]` check is working right.  This typetest line then results in a `param_0: {}` parameter being passed onwards to the next selectors in the chain, which causes an error a couple selectors later:

```ts
  const selector8: Selector<State, string> = createSelector(selector7, s => s)
  ```

Also, `createStructuredSelector` is now unable to infer some of its types correctly[.

Perf-wise, this seems to be giving "excessively deep" errors around selector 10.  I tried going back to 2a314fd , which supposedly got things up to 15 at once per #534, but that also seems to be dying around 10 deep, so I'm confused where the improvement supposedly happened.

The lockfile rewrite is because I finally bumped myself to `npm@8` recently and then happened to install `ts-toolbelt`.  Not thrilled about that, but we'll sort it out later.

@phryneas , @eXamadeus : would appreciate any further insight you have on this one!


eXamadeus:
Just commenting here because my week was crazy busy. I'll make some comments here over the weekend.

markerikson:
Yeah, I'm sitting down to see if I can pick up where I left off on this.

markerikson:
Ended up spending yesterday on https://github.com/reduxjs/redux-toolkit/pull/1730 , but actually looking at this now.

markerikson:
The latest commit  ( 573455e) attempts to rewrite the `MergeParameters` type to drop all of the `ts-toolbelt` usage I had in an earlier commit.

It works. Sorta. 

It does actually spit out the right results in the "playground" file I added:

image

but, there's a TS "can't index this" error in the type itself, and if I use it for real, I also get type failures with the actual usages of `MergeParameters` elsewhere in `types.ts`

markerikson:
This WIP PR is superseded by #549 .

Issue #543, opened 2021-11-08
Build of library with declarations: true


abelgibayev:
Hello! 

I need to build a core library for multiple applications with common slices and selectors inside. To have a good typization of selectors I need to set `declarations: true` in `tsconfig.json`. But it doesn't work, because OutputSelectorFields is not exported. The error I get from TS is `Exported variable 'selectAccountUser' has or is using name 'OutputSelectorFields' from external module "path-to-folder/node_modules/reselect/es/types" but cannot be named.`

Reselect version 4.1.2

JamesGelok:
Same here. 

Downgrading to 4.1.1 was my only option:

```json
  "dependencies": {
    "...": "",
    "reselect": "4.1.1",
    "...": "",
  },
```

markerikson:
I'll try to look at this tomorrow. I'm not 100% sure that I actually want to export the `OutputSelectorFields` type as the solution here, as I've been looking at trying to rearrange the typedefs some.

onuralpbicer:
Having the same issue

markerikson:
Could someone put together a repo that demonstrates this problem happening?  ie, something with a "shared selector" lib and a consuming lib.  That would help me investigate the problem and try to make sure it's resolved correctly.

Methuselah96:
This isn't a small repo, but I had to work around it in [this PR](https://github.com/reduxjs/redux-devtools/pull/951) in `redux-devtools`.

markerikson:
Fixed in https://github.com/reduxjs/reselect/releases/tag/v4.1.3

Issue #541, opened 2021-11-05
Type checking broken in 4.1


ViktorQvarfordt:
In this example, reselect@4.0 correctly gives a typescript error for the last line, while reselect@4.1 reports no error

```ts
const selectProp1 = createSelector(
  [state => state, (state, props: { prop1: number }) => props],
  (state, { prop1 }) => [state, prop1]
)

const selectProp2 = createSelector(
  [selectProp1, (state, props: { prop2: number }) => props],
  (state, { prop2 }) => [state, prop2]
)

selectProp1('state', { prop1: 1 })
selectProp2('state', { prop2: 2 })
```

Here is a screenshot of the error, for reference:

![image](https://user-images.githubusercontent.com/344809/140572997-3dc275b5-ad2a-4840-8b94-92f053497973.png)

There are plenty other typing errors in 4.1 already reported in issue https://github.com/reduxjs/reselect/issues/540.

markerikson:
Fixed in https://github.com/reduxjs/reselect/releases/tag/v4.1.3

PR #545, opened 2021-11-09
fix(): export OutputSelectorFields

chyzwar:
When trying to export selector from shared package we are getting errors. 

```
Exported variable 'xxxxxx' has or is using name 'OutputSelectorFields' from external module "/home/vagrant/Projects/common-2/node_modules/reselect/es/types"
```

Fix is to export OutputSelectorFields


chyzwar:
It would be nice to get this merged, reselect managed to break our code twice, 4.1.1 and 4.1.2 releases raised TS errors. Because we share selectors across multiple repositories we need to upgrade this in sync otherwise repo with 4.1.0 will throw errors when consuming selectors in version 4.1.1. We cannot upgrade to 4.1.2 because of above error.

andymikulski:
I've also hit this issue, and as a temporary measure have resorted to just editing my local `node_modules` in order to continue work.

chyzwar:
@markerikson Any chance you can take a look at this ?

markerikson:
I've spent my entire last two weekends trying to come up with solutions for #540 and #541.  As of yesterday I _think_ I have a resolution, but it's going to take some additional time to get that work cleaned up and ready.

Once that's done I'm going to look at the `OutputSelectorFields` issue. Per #543 , I may end up actually removing that type, but what I really need is for someone to provide me a repo that demonstrates this problem happening so that I can reproduce the behavior locally and confirm that any fixes work.

So, I understand the concerns, and I'm trying to address them, but the best thing you can do to help is provide a repro for me.

Methuselah96:
There's a repro in redux-devtools if that helps (linked in [the issue](https://github.com/reduxjs/reselect/issues/543)). I only had problems when I had to manually set the generics for `createSelector`.

markerikson:
@Methuselah96 can you narrow it down to a smaller standalone repo?

markerikson:
Hmm. Just an import, not even a "re-export from a shared lib"?

Methuselah96:
Right, the issue is creating a .d.ts file since it can't access `OutputSelectorFields`.

markerikson:
Fixed in https://github.com/reduxjs/reselect/releases/tag/v4.1.3

PR #549, opened 2021-11-16
Rewrite function parameter type inference to fix assorted issues

### LINKED ISSUE #540: Multiple selectors with arguments result in a union type

### LINKED ISSUE #541: Type checking broken in 4.1

### LINKED ISSUE #543: Build of library with declarations: true

### LINKED ISSUE #546: Investigate adding functionality from other selector-related libs
See https://github.com/reduxjs/reselect/discussions/491 for a list of other selector libs to compare with.  

I think the debugger tools from https://github.com/skortchmark9/reselect-tools would be a top candidate.

Also, https://github.com/sgrishchenko/reselect-utils has some interesting utilities for various use cases.

markerikson:
This PR:

- Adds several additional typetests based on reported issues
- Rewrites the types for extracting input selector parameters for correctness
- Expands the displayed type of output selectors to simplify view of the fields added by the memoize function (ie, `{clearCache: () => void}`)
- Fixes issues with `createStructuredSelector` not inferring the types of input arguments
- Exports the `OutputSelectorFields` type to fix breakages with declaration generation

Fixes #540
Fixes #541 
Fixes #543 
Fixes #546

This completely rewrites the implementation of the `MergeParameters` type, and _should_ fix all correctness issues with selector input parameter behavior.

The actual implementation is a doozy. It's a combination of multiple ideas and techniques from myself, @phryneas , @Retsam, @pfgray, and umpteen dozen Stack Overflow answers and TS repo discussions.

Loosely summarized, the behavior is:

- Extract the "type of the parameters" from each input selector, and get them all in a nested tuple
- Transpose the tuple so that we now have `[ allIndex0Args, allIndex1Args, ....]`
- Intersect all args at a given index to determine exactly what is legal to pass in

I also updated `createStructuredSelector` to do a better job of inferring all of its parameters, and exported the `OutputSelectorFields` type because TS was failing to export any selector variables because it didn't have access to that.  (I suspect I could have eliminated the need for it if I'd done a full `ComputeDeep` on the selector definition, but oh well.)


markerikson:
This passes, I checked that the use of `OutputSelectorFields` is fixed locally in a test project, and this fixes a bunch of issues. Let's just get this out there.

PR #553, opened 2021-11-17
fix(): fix re-export of createSelectorCreator

chyzwar:
Fix type error when trying to re-export createSelectorCreator.
Added simple test, I am not sure if you want to have type test together with unit tests


chyzwar:
@markerikson if you have time to take a look at above.

markerikson:
Ended up doing this separately in #557 , thanks!

PR #556, opened 2021-11-24
Fix intersection of parameters that are undefined

markerikson:
This PR:

- Updates the input selector parameter inference to correctly handle cases where a parameter is `undefined` or `null` as opposed to unions that contain `undefined` at all.

This should now work much better for functions that have optional parameters, as well as functions with mismatched input selector parameter lengths.


Retsam (commented):
LGTM

Issue #563, opened 2022-01-19
Type inference issues caused by default param values or any/unknown type (v4.1.5)


hiredgun:
I encountered two issues with type inference in v4.1.5:

1. When an input-selector expects more than one param and the last param is typed as unknown/any, TS complains that the final selector expect 1 param instead of 2.

2. When a default value is provided to any param of at least one input-selector without providing a type explicitly, all params to an output-selector are implicitly typed as any. 

This code sandbox shows the issue: https://codesandbox.io/s/reselect-ts-issues-opbhs?file=/src/index.ts:272-279

markerikson:
Hmm. Afraid I don't have time to investigate this atm, but my gut says there may not be much we can do about either of these.

We get the types of the function arguments using the TS `Parameters` built-in type.  If it isn't including those arguments in the result, I'm not sure there's any way for us to adjust for that.

lukeapage:
It seems to be a general issue that if the infer fails for the parameters, it falls back to assume there is no argument and hides any error e.g. the 2nd issue, if I change it to this

```
const getDefaultKey = (state: State) => state.defaultKey;

const selector2 = createSelector(
  (state: State, key: string) => "key",
  (state: State, key2: number) => key2,
  (state: State, key3: string) => key3,
  (key, key2, key3) => key + key2 + key3
);

selector2({} as State); // No error

selector2({} as State, 2); // error, but unexpected
```

as a consumer of the selector I am being told by typescript to not give a 2nd argument and if I don't I get no error. but my selectors actually *require* a 2nd argument but are mistyped, maybe one is a number and should have been a string.

lukeapage:
The problem is that

```
/** Determines the combined  "Params" type (all remaining args) from all input selectors */
export type GetParamsFromSelectors<
  S extends SelectorArray,
  RemainingItems extends readonly unknown[] = Tail<MergeParameters<S>>
  // This seems to default to an array containing an empty object, which is
  // not meaningful and causes problems with the `Selector/OutputSelector` types.
  // Force it to have a meaningful value, or cancel it out.
> = RemainingItems extends [EmptyObject] ? never : RemainingItems
```

`RemainingItems` can be `[never]` and it seems that `[never]` ends up adding onto a params array and just becoming an empty array.

We can probably escape out to never if remaining items is `[never]` though to support multi arguments we would need a way to determine if the array can contain never

I am having a go at a fix :S

lukeapage:
[...]

PR #564, opened 2022-01-20
Make it an error if optional types are inconsistent

lukeapage:
Improves things for #563

Honestly this PR is mostly to demonstrate the problem and what could be done.. I tried more fundamental changes but someone who knows the codebase better would need to try. I didn't understand why Selectors default to `never` and then switch between having args or not. as per this playground

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAFASiQPhgbwL4CwAUKJLAEZKoB0tcATgOYQBcMA2gLoaLb7HE4KAORQAptGFoA3MXIjxkmUA

if the args are typed as empty tuple e.g.  `[]` then typescript will enforce that no params are passed. Can't we default to empty arrays?

I believe that would be the only change required and would probably reduce complexity and code as a tuple containing never will not allow 0 or 1 arguments.

https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDAFASiQPhgbwL4CwAUKJLAEZKoB0tcATgOYQBcMA2mAKYBuX9AXQyJs+YsTgoA5FC7QpaANzFy02fKUr0ioA

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -92,10 +92,7 @@ export type GetStateFromSelectors<S extends SelectorArray> =
 export type GetParamsFromSelectors<
   S extends SelectorArray,
   RemainingItems extends readonly unknown[] = Tail<MergeParameters<S>>
-  // This seems to default to an array containing an empty object, which is
-  // not meaningful and causes problems with the `Selector/OutputSelector` types.
-  // Force it to have a meaningful value, or cancel it out.
-> = RemainingItems extends [EmptyObject] ? never : RemainingItems
+> = RemainingItems
 
 /** Given a set of input selectors, extracts the intersected parameters to determine
  * what values can actually be passed to all of the input selectors at once
--- a/typescript_test/test.ts
+++ b/typescript_test/test.ts
@@ -610,9 +610,8 @@ function testOptionalArgumentsConflicting() {
     (str) => str
   )
 
-  // @ts-expect-error This should be an error because prefix is required, although it is any
+  // @ts-expect-error
   selector2({} as State)
-  // @ts-expect-error Ideally this is not an error, so this expect-error is a bug
   selector2({} as State, 'blach')
 
   const selector3 = createSelector(
--- a/typescript_test/test.ts
+++ b/typescript_test/test.ts
@@ -164,8 +164,12 @@ declare function connect<S, P, R>(
   selector: ParametricSelector<S, P, R>
 ): (component: Component<P & R>) => Component<P>
 
+declare function connect2<S, P, R>(
+  selector: Selector<S, P>
+): (component: Component<P & R>) => Component<P>
+
 function testConnect() {
-  connect(
+  connect2(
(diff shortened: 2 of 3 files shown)


typescript_test/test.ts
```diff
@@ -164,8 +164,12 @@ declare function connect<S, P, R>(
   selector: ParametricSelector<S, P, R>
 ): (component: Component<P & R>) => Component<P>
 
+declare function connect2<S, P, R>(
```
lukeapage, line 167: I couldn't get the existing connect arguments to work.

using ParametricSelector implies there is a parameter, but there is not... and this then disagrees with an implied type of having no parameter... sorry I didn't know what to do here and it might be a reason why a different fix is needed.

typescript_test/test.ts
```diff
@@ -580,6 +584,46 @@ function testArrayArgument() {
...
+  selector({} as State, 'string')
+  // @ts-expect-error
+  selector({} as State, 1)
+
+  const selector2 = createSelector(
+    (state: State, prefix: any) => prefix + state.foo,
+    (str) => str
+  )
+
+  // @ts-expect-error This should be an error because prefix is required, although it is any
```
lukeapage, line 613: so this doesn't fix problem 1 with optional params, for problem 2 it makes not passing an argument an error, which is good as the current typings mislead the user into not passing a param.

It does solve problem 3 that inconsistent typings end up just eliminating the optional param.

src/types.ts
```diff
@@ -92,10 +92,7 @@ export type GetStateFromSelectors<S extends SelectorArray> =
 export type GetParamsFromSelectors<
   S extends SelectorArray,
   RemainingItems extends readonly unknown[] = Tail<MergeParameters<S>>
-  // This seems to default to an array containing an empty object, which is
-  // not meaningful and causes problems with the `Selector/OutputSelector` types.
-  // Force it to have a meaningful value, or cancel it out.
-> = RemainingItems extends [EmptyObject] ? never : RemainingItems
+> = RemainingItems
```
lukeapage, line 95: As commented on the issue, changing this little bit fixes issues and does not break any other typescript test..

lukeapage:
@markerikson would you have time to let me know what I can do to get this merged ?

we haven’t upgraded to the new typings, so I can put some effort into this, but I need to know what to do..

markerikson:
Honestly, I'm juggling enough things in my head atm that I don't know what's going on with this PR :)  

Does this actually make the problems in the original issue go away?  (not seeing any test runs here for some reason - maybe because it's an external PR?)

lukeapage:
Hi @markerikson,

I appreciate the reply, yes it does fix 2 out of 3 issues.

how can I make it easiest for you? 

Maybe I will split the changes so it’s clear what it changes and what it fixes ?

lukeapage:
superceded by https://github.com/reduxjs/reselect/pull/572

PR #572, opened 2022-04-08
Fix #563 - Make it an error when the typings become incorrect

### LINKED ISSUE #563: Type inference issues caused by default param values or any/unknown type (v4.1.5)

lukeapage:
I've taken my Pr - https://github.com/reduxjs/reselect/pull/564 and simplified the change to only remove the condition that breaks. 

It is incorrect because when a problem happens with params, `[never]` extends `[EmptyObject]` and `RemainingItems` becomes `[never]`. 

But when we get to OutputSelector
```
export type OutputSelector<
  S extends SelectorArray,
  Result,
  Combiner extends UnknownFunction,
  Params extends readonly any[] = never // MergeParameters<S>
> = Selector<GetStateFromSelectors<S>, Result, Params> &
  OutputSelectorFields<Combiner>
```

never is the default to mean that there should not be any params (e.g. no problem at all, rather than [never] which means there is a type problem - there is 1 parameter of value never, which then will error with no params and error with 1 param).

The alternative fix requires alot of changes and this simple fix does not break any existing test cases.

@markerikson I will comment on the Pr which pieces of the test case I added are different on master.

I hope this new Pr is easier to review.

### DIFF
--- a/typescript_test/test.ts
+++ b/typescript_test/test.ts
@@ -580,6 +580,104 @@ function testArrayArgument() {
   }
 }
 
+function testOptionalArgumentsConflicting() {
+  type State = { foo: string; bar: number; baz: boolean }
+
+  const selector = createSelector(
+    (state: State) => state.baz,
+    (state: State, arg: string) => arg,
+    (state: State, arg: number) => arg,
+    (baz) => {
+      const baz1: boolean = baz
+      // @ts-expect-error
+      const baz2: number = baz
+    }
+  )
+
+  // @ts-expect-error the selector above has inconsistent conflicting arguments so usage should error
+  selector({} as State)
+  // @ts-expect-error
+  selector({} as State, 'string')
+  // @ts-expect-error
+  selector({} as State, 1)
+
+  const selector2 = createSelector(
+    (state: State, prefix: any) => prefix + state.foo,
+    (str) => str
+  )
+
+  // @ts-expect-error here we require one argument which can be anything so error if there are no arguments
+  selector2({} as State)
+  // no error passing anything in
+  selector2({} as State, 'blach')
+  selector2({} as State, 1)
+
+  // here the argument is optional so it should be possible to omit the argument or pass anything
+  const selector3 = createSelector(
+    (state: State, prefix?: any) => prefix + state.foo,
+    (str) => str
+  )
+
+  selector3({} as State)
+  selector3({} as State, 1)
+  selector3({} as State, 'blach')
+
+  // https://github.com/reduxjs/reselect/issues/563
+  const selector4 = createSelector(
+    (state: State, prefix: string, suffix: any) => prefix + state.foo + String(suffix),
+    (str) => str
+  )
+
+  // @ts-expect-error
+  selector4({} as State)
+  // @ts-expect-error
+  selector4({} as State, 'blach')
+  selector4({} as State, 'blach', 4)
+
+  // as above but a unknown 2nd argument
+  const selector5 = createSelector(
+    (state: State, prefix: string, suffix: unknown) => prefix + state.foo + String(suffix),
+    (str) => str
+  )
+
+  // @ts-expect-error
+  selector5({} as State)
+  // @ts-expect-error
+  selector5({} as State, 'blach')
+  selector5({} as State, 'blach', 4)
+
+  // @ts-expect-error It would be great to delete this, it is not correct.
+  // Due to what must be a TS bug? if the default parameter is used, we lose the type for prefix
+  // and it is impossible to type the selector without typing prefix
+  const selector6 = createSelector(
+    (state: State, prefix = '') => prefix + state.foo,
+    (str: string) => str
+  )
+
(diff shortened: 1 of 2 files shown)


typescript_test/test.ts
```diff
@@ -580,6 +580,104 @@ function testArrayArgument() {
...
+    (state: State, arg: string) => arg,
+    (state: State, arg: number) => arg,
+    (baz) => {
+      const baz1: boolean = baz
+      // @ts-expect-error
+      const baz2: number = baz
+    }
+  )
+
+  // @ts-expect-error the selector above has inconsistent conflicting arguments so usage should error
```
lukeapage, line 597: on master there is no error - even though we expect a parameter, we are allowed to call it without one

typescript_test/test.ts
```diff
@@ -580,6 +580,104 @@ function testArrayArgument() {
...
+  // @ts-expect-error
+  selector({} as State, 1)
+
+  const selector2 = createSelector(
+    (state: State, prefix: any) => prefix + state.foo,
+    (str) => str
+  )
+
+  // @ts-expect-error here we require one argument which can be anything so error if there are no arguments
+  selector2({} as State)
```
lukeapage, line 610: on master there is no error, again, we require a parameter but none is passed in

typescript_test/test.ts
```diff
@@ -580,6 +580,115 @@ function testArrayArgument() {
...
+  selector7({} as State, 'blach')
+  // @ts-expect-error wrong type
+  selector7({} as State, 1)
+
+  const selector8 = createSelector(
+    (state: State, prefix: unknown) => prefix,
+    (str) => str
+  )
+
+  // @ts-expect-error needs a argument
```
lukeapage, line 685: this error does not occur on master

markerikson:
Thanks, I'll _try_ to take a look at this in the next few days, but also have some other things I need to get done first.

markerikson:
For the record I haven't forgotten about this :)  But I've been really busy with a bunch of other things and haven't had time to do general Redux issue-related work.

Still high on my list to look at as soon as I have time...

lukeapage:
Thanks @markerikson I have been watching the redux / toolkit releases and figures as much, I’m a burnt out open source maintainer myself so I understand.

typescript next failed but I suspect master is broken, will try to take a look soon.

markerikson:
I think the TS error there is actually related to something on the RTK side - we've been seeing the same errors in the RTK repo, I think.

markerikson:
Okay, this seems reasonable.  I'll merge and see if there's any other tweaks worth doing before publishing 4.1.6

markerikson:
Hmm. So I opened this up in VS Code Web and didn't immediately see errors, so I merged it. But, when I pulled `master` locally and opened up `types.ts`, I'm seeing a couple TS errors in the changes you added, using TS 4.5.  So, I'm kinda confused.  Can you review the changes on your side?  

Specifically, I'm seeing errors with:

```ts
function testConnect() {
  connect(
    createSelector(
      (state: { foo: string }) => state.foo,
      foo => ({ foo })
    )
  )
```

and:

```ts
  // ------  ISSUE 1 ------

    (state: State, prefix: unknown) => prefix + state.text,
    str => str //          ^ unknown or any cause the error below
  )

  selector(exampleState, 'blach')
```

lukeapage:
I can check tomorrow, it’s late here in Denmark

lukeapage:
Hi @markerikson,

I've tried TS 4.4, 4.5 and 4.6 - for each I reloaded my editor (webstorm) and waited for TS errors and also ran `yarn test:typescript` and I cannot see the problem you are talking about.. Can you give me more info and check it is not a local issue for you?

markerikson:
Whoops, I apologize!  I had apparently doing a bit of experimenting the last time I touched the Reselect repo several weeks ago, and had some changes on disk that weren't committed and weren't relevant.  When I did a `git pull`, those caused conflicts and the pull didn't get checked out, so I was actually looking at the same uncommitted code I had already.

If I ditch those changes, the file looks clean.

lukeapage:
@markerikson any chance we can have a new release to be able to consume this? I'm happy to be tagged if it results in issues for others when they consume it.

markerikson:
Yeah, sorry, been busy with a bunch of other things.  I'll try to push out a new build this weekend.

markerikson:
@lukeapage Okay, this is _finally_ live in https://github.com/reduxjs/reselect/releases/tag/v4.1.6 . Thanks for the work and the nudges here!

(And if you have any other ideas for improving the types, please feel free to file more PRs - always interested in making things better!)

PR #573, opened 2022-04-27
Add a failed test and naive fix for issue 571

chentsulin:
Added a failed case for issue 571.

I can confirm this test failed on my machine, but still can't figure out how to fix it.


markerikson:
@chentsulin : can you clarify how that tweak to the types is intended to fix this issue?

PR #587, opened 2022-11-03
TS 4.9 compat

markerikson:
- Added TS 4.7, 4.8, and 4.9 RC to the test matrix
- Bumped the in-repo TS version to 4.8
- Reconfigured all the types so that TS 4.6 and earlier get the existing `MergeParameters` implementation (which actually worked up through 4.8), whereas 4.7+ get the new implementation from https://github.com/microsoft/TypeScript/pull/50831#issuecomment-1253830522
  - This required stupid hacky build-time package shenanigans that we previously did over in the Redux Toolkit repo to handle TS <4.1.  We removed those changes in https://github.com/reduxjs/redux-toolkit/pull/2249 in preparation for RTK 1.9 - see the _removed_ code in that PR

Pasting the main comment from the big commit:

> Per https://github.com/microsoft/TypeScript/pull/50831 , the existing implementation of `MergeParameters` broke with TS 49 due to a change in how "optional" and "undefined" arguments get interpreted
> 
> Anders Hjelsberg himself supplied a new implementation that's actually much simpler, but only works with TS 4.7 and greater.
> 
> Normally this would require shipping two entirely different sets of
types, as we already do for TS <4.1. However, there's a trick we used
with RTK where we:
> 
> - Create a folder and put two different files with different impls of the same type inside
> - Add an index file and re-export one of those
> - Add a file named `package.dist.json` containing `typesVersions` that points to both of those `.d.ts` files
> - Copy that `package.dist.json` over to `dist/some/package.json` during the actual build/publish step
> 
> That way during dev TS just imports the type as normal, but the built
version sees that `some/package.json`, sees `typesVersions`, finds the
right `.d.ts` file, and imports the right implementation for itself.


markerikson:
Confirmed this works by:

- Creating a new Vite React+TS app project
- Copying the _entire_ typetest file into that
- doing a `yalc publish` from this branch and installing into that repo
- taking turns installing TS 4.2, 4.6, 4.8, and 4.9-rc and running `yarn tsc` to confirm that the typetest file compiles cleanly
- uncommenting the dead chunk around like 615 where TS 4.8 and earlier had a `// @ts-expect-error`, and confirming that 4.9 panics because there _isn't_ an error (also confirming that they all really were compiling the file)

PR #592, opened 2022-12-07
fix types on resultFunc (#591)

### LINKED ISSUE #591: incorrect types: clearCache getting added to resultFunc return value
I think there are incorrect types for the `resultFunc` property that you can pull off of a created selector.

```ts
function isOdd(num: number): boolean {
  return num % 2 === 1;
}

const memoizedIsOdd = createSelector([isOdd], (bool: boolean) => bool);

memoizedIsOdd.resultFunc(true).clearCache(); //doesn't exist, but the types think it does
```

clearCache is only on the actual selector returned by createSelector, not the return value of result func. The types think it's on both though.

I am on version 4.1.7

bdwain:
Fixes #591 

Happy to make tweaks to this, but I think it's roughly what is needed to fix the broken types

From original issue.

I think there are incorrect types for the `resultFunc` property that you can pull off of a created selector. I am on version 4.1.7

```ts
function isOdd(num: number): boolean {
  return num % 2 === 1;
}

const memoizedIsOdd = createSelector([isOdd], (bool: boolean) => bool);

memoizedIsOdd.resultFunc(true).clearCache(); //doesn't exist, but the types think it does
```

clearCache is only on the actual selector returned by createSelector, not the return value of result func. The types think it's on both though.

node output:
```
Welcome to Node.js v16.18.1.
Type ".help" for more information.
> function isOdd(num) {
...   return num % 2 === 1;
... }
undefined
>
> const { createSelector } = require('reselect')
undefined
> createSelector
[Function: createSelector]
> const memoizedIsOdd = createSelector([isOdd], (bool) => bool);
undefined
> memoizedIsOdd.resultFunc.clearCache
undefined
> memoizedIsOdd.clearCache
[Function (anonymous)]
> memoizedIsOdd.resultFunc(true)
true
> memoizedIsOdd.resultFunc(true).clearCache
undefined
> memoizedIsOdd.resultFunc(true).clearCache()
Uncaught TypeError: memoizedIsOdd.resultFunc(...).clearCache is not a function
>
```


src/index.ts
```diff
@@ -176,8 +176,9 @@ export interface CreateSelectorFunction<
   ): OutputSelector<
     Selectors,
     Result,
-    (...args: SelectorResultArray<Selectors>) => Result & Keys,
-    GetParamsFromSelectors<Selectors>
+    (...args: SelectorResultArray<Selectors>) => Result,
```
bdwain, line 179: I think this was meant to be `((...args: SelectorResultArray<Selectors>) => Result) & Keys`

which would have been closer to working, but it would have put clearCache on resultFunc, which it wasn't actually on, so I needed to pass in the Keys as their own property to be able to add them only to memoizedResultFunc

markerikson:
Seems reasonable at first glance. Thanks!

I'll try to publish a new version no later than this weekend.

bdwain:
thanks!

Issue #601, opened 2023-03-14
Rewrite of MergeParameters in v4.1.7 breaks type inference with spreading in createSelector


PeterBul:
Before version 4.1.7 and the rewrite of MergeParameters, it was possible to write 
```ts
interface ITestState {
  foo: number;
  bar: string;
}

const selectFoo = (state: ITestState) => state.foo;
const selectBar = (state: ITestState) => state.bar;

const selectors = [selectFoo, selectBar];

export const selectFooBar = createSelector(
  [selectFoo, selectBar, ...selectors],
  (foo, bar, ...rest) => {
   ...
  }
);
```
This would infer the type of `newSelector` to be `(state: ITestState) => ... & ...` since all selectors used as input had that type.

In 4.1.7, after the rewrite of MergeParameters, `newSelector` is of type `(state: never) => ...`

It still works with only using
```ts
export const selectFooBar = createSelector(
  [selectFoo, selectBar],
  (foo, bar) => {
   ...
  }
);
```

or

```ts
export const selectFooBar = createSelector(
  [...selectors],
  (...rest) => {
   ...
  }
);
```

but the combination does not work any longer.

Here is the example on [StackBlitz](https://stackblitz.com/edit/typescript-19vlgn?file=index.ts).

PeterBul:
Say you want to add a series of input selectors to an array based on some data. Then you don't know how your selectors will look, but writing it like this will enable you to get an array of results which you then can use in the new selector.

markerikson:
Huh. Definitely not a pattern I've ever seen in use.

To be honest this is going to be very low priority - there's a lot of other things going on atm, and I don't know if/when I'll have time to look at this.  If you or anyone else has a suggested fix and can add a typetest showing the improved type works correctly, that'd be appreciated!

PeterBul:
I can see if I can figure it out 👍

nicolaschambrier:
There might be a more general semver issue here. We recently had changes to apply to upgrade from a 4.x to 4.1.8, and we were surprised it would happen for a minor version.

When exploring a bit we found out the signatures changed a lot through minor versions, in a breaking way for the build step:

```ts
// 4.0.0
export function createSelector<S1, R1, T>(
  selectors: [Selector<S1, R1>],
  combiner: (res1: R1) => T,
): OutputSelector<S1, T, (res1: R1) => T>;

// 4.1.8
<Selectors extends SelectorArray, Result>(
    selectors: [...Selectors],
[...]

PR #602, opened 2023-03-15
Infer correct state when input selectors are a mix of explicit and rest syntax in createSelector

PeterBul:
The state parameter of the result of createSelector was of type never if input selectors were a mix of explicit and spread selectors. Additionally, the type of extra parameters are lost if the input has this format.

This PR aims to allow the combining of explicitly defined input selectors and using rest syntax in the end. It does so by making sure that the type of the array that is part of the rest syntax is transformed into a normal type before merging the parameter tuples. This is valid since we really don't care whether we have 1 or 10 selectors in the array that is spread as long as the types are the same.

This is related to [issue 601](https://github.com/reduxjs/reselect/issues/601) where some more details can be found in the conversation.

### DIFF
--- a/src/versionedTypes/ts47-mergeParameters.ts
+++ b/src/versionedTypes/ts47-mergeParameters.ts
@@ -34,11 +34,9 @@ type MergeTuples<
   [K in keyof L]: Intersect<ElementsAt<T, K>>
 }
 
-type ExtractParameters<T extends readonly UnknownFunction[]> = {
+export type ExtractParams<T extends readonly UnknownFunction[]> = {
   [K in keyof T]: Parameters<T[K]>
 }
 
 export type MergeParameters<T extends readonly UnknownFunction[]> =
-  '0' extends keyof T
-    ? MergeTuples<ExtractParameters<T>>
-    : Parameters<T[number]>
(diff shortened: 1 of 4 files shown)


src/versionedTypes/ts47-mergeParameters.ts
```diff
@@ -1,23 +1,21 @@
 // This entire implementation courtesy of Anders Hjelsberg:
 // https://github.com/microsoft/TypeScript/pull/50831#issuecomment-1253830522
 
+import { ReverseHead, ReverseTail } from '../types'
+
 type UnknownFunction = (...args: any[]) => any
 
-type LongestTuple<T extends readonly unknown[][]> = T extends [
```
PeterBul, line 6: The `readonly unknown[][] `requirement has to be removed from several places since `ReverseHead` and `ReverseTail` use recursive types to find the first an last tuple element. Typescript complains when a condition (like `readonly unknown[][]`) has to to be compared with a type that is recursive, since it can't check the result of that type.

The checks of whether a type is `readonly unknown[][]` are moved to conditintional types instead where it is needed. 

These are internal types specifically used for this use case, so removing the requirement should not be an issue.

Issue #298, opened 2017-11-11
Can't set selector memoization function


Mosho1:
from `index.js:68`:

```
const selector = defaultMemoize(function () {
      const params = []
      const length = dependencies.length

      for (let i = 0; i < length; i++) {
        // apply arguments instead of spreading and mutate a local list of params for performance.
        params.push(dependencies[i].apply(null, arguments))
      }

      // apply arguments instead of spreading for performance.
      return memoizedResultFunc.apply(null, params)
    })
```

Why should this not use the provided `memoize`? All tests pass with the change, but any other option to allow setting this will be very helpful. another helpful addition is passing `memoizeOptions` here as well, but this does break a test case which will have to be updated, although I do believe this is the expected behavior as an end user.

https://github.com/reactjs/reselect/pull/297

ellbee:
Will be addressed in version 4

Issue #279, opened 2017-08-28
[Enhancement] Selector Debugging Tools


skortchmark9:
Firstly, I love reselect, it's possibly my favorite part of the redux ecosystem. I love it so much in fact, that I could be accused of overusing selectors - I like to shove as much logic as possible into them, so that my teammates and I can share computed data in a performant way across our app.

Because we rely so heavily on selectors for our logic, selectors are often where our bugs crop up. Redux's debugging tools are excellent, and I'm imagining reselect could pair its debugging tools nicely along with them. A particular emphasis is **"No Print Lines"**. I like that all the information about my app's state store is already present for me to view without a code change. If all the information about the app's *computed* state were available also, I would never have to change my code to figure out what was broken.

# Ideas
## 1.  Graph
The first idea is essentially the same as what was originally proposed in #71 : a visual dependency graph of selector parents and children:
![image](https://user-images.githubusercontent.com/6496811/29755114-7519eb88-8b47-11e7-86d9-80d3eadb0f0d.png)
(picture grabbed from #71 , originally from @MattSPalmer)
This would be really useful for refactoring heavily depended-on selectors, as well as debugging.
## 2. Click-to-select
The ability to click a selector on that graph to see what its inputs / outputs are at a given state.

I don't really like the logging solutions in #71 because they don't let you surgically inspect a particular selector without making code changes. Same issue with @ajwhite 's https://github.com/ajwhite/reselect-devtools.

# Implementation
I don't think it would be too hard to make a drop-in wrapper for ```createSelector``` which populates the dependency graph:
```javascript
function createSelector(...dependencies) {
  const resultFunc = dependencies.pop();
  const selector = Reselect.createSelector(...dependencies, resultFunc);
  selector.dependencies = dependencies;
  Redebug.selectorGraph[selector] = selector;
  return selector;
}
```
And click-to-select just needs ```getState()``` injected, either via middleware or something even more vanilla:
```javascript
Redebug.getStateWith(() => store.getState());
```

# Challenges
## Naming
The one thing I am struggling with is getting the selector names. They're usually anonymous! In my ```createSelector``` wrapper, I'm keying the graph based on the ```resultFunc```'s ```toString()```, which is going to result in pretty ugly display names. For example:
```javascript
const currentUser$ = createSelector(ui$, users$, (ui, users) => users[ui.currentUser]);
=> "(ui, users) => users[ui.currentUser]"
```
[...]

PR #605, opened 2023-04-12
Add experimental new memoizers: autotrack and weakmap

markerikson:
This PR:

- Switches the memoization of incoming arguments back to `defaultMemoize` for perf atm
  - **NOTE**: this will require significantly more investigation and consideration. See: 
    - #279 
    - #297 (the actual change that put `memoize()` there now)
    - #298 
    - #359 
- Adds 3 new experimental memoizers:
  - `autotrackMemoizer`: uses Proxies and an "autotracking" implementation to determine _exactly_ which fields are accessed in a selector, and only recalculate when those fields get changed
  - `weakMapMemoizer`: uses `WeakMap`s to form a tree of cache lookups based on args
  - ~~`signalisMemoizer`: uses https://github.com/cafreeman/signalis to try to do something similar to the `autotrack` approach~~

There's a good chance that these could go into v4.x, but I'm putting them in a 5.x-based build branch for now.

### Status and Tradeoffs

Currently, `autotrack` and `weakMap` seem to actually work and pass tests. `signalis` runs but doesn't memoize properly.

`autotrack` works well! I can confirm that if you do something like `todos => todos.map(t => t.id)`, and then flip `t.completed`, it will _not_ recalculate.  However, it also is definitely slower than `defaultMemoize`.  The question is whether the slowness is enough to matter in practice.  On the flip side, the fact that it can recalculate a lot less often in some cases means that there'd be fewer re-renders.

`weakMap` is closer in speed to `defaultMemoize`. Haven't done significant testing on it yet. Seems like there's some amount of multi-size behavior by default, but haven't figured out how that grows over time or gets GC'd.

### Resources / inspiration / places I swiped code from:

- Autotrack:
  - https://github.com/simonihmig/tracked-redux
  - https://www.pzuraq.com/blog/how-autotracking-works
  - https://gist.github.com/pzuraq/79bf862e0f8cd9521b79c4b6eccdc4f9
  - https://github.com/lifeart/glimmerx-workshop
  - https://v5.chriskrycho.com/journal/autotracking-elegant-dx-via-cutting-edge-cs/
  - https://github.com/glimmerjs/glimmer-vm/blob/master/packages/%2540glimmer/validator/
- WeakMap:
  - Straight out of the React Cache source: [`ReactCache.ts`](https://github.com/facebook/react/blob/0b974418c9a56f6c560298560265dcf4b65784bc/packages/react/src/ReactCache.js#L55-L126)
- Signalis:
  - https://github.com/cafreeman/signalis/issues/53


markerikson:
Aaaand the build is totally busted due to some Reselect -> RTK -> Reselect dependency thing.

Will worry about that later.

markerikson:
Updates since the original PR creation:

- Deleted the `signalisMemoize` implementation. It didn't work right.
- Cleaned up `autotrackMemoize` and `weakmapMemoize`
- Did some testing on `weakmapMemoize` in conjunction with @crutchcorn and concluded that `weakmapMemoize` has _effectively_ an infinite cache size, in that the use of `WeakMap`s will keep around cached values as long as the args are still in memory, and then clean up the cached values as soon as those args are GC'd.
- Split the unit tests into several files just for readability, and to enable adding copy-pastes of the "basic selector" tests using both `autotrackMemoize` and `weakmapMemoize`

PR #612, opened 2023-05-13
Run input selectors twice, to check stability.

### LINKED ISSUE #611: Run selectors twice in development to warn for memoization failures
This one has been asked about for a while, particularly by @theKashey :

- https://github.com/reduxjs/reselect/discussions/491#discussioncomment-5765670
- https://twitter.com/keeslinp/status/1656342003268071424

The idea is that users sometimes end up writing badly-written selectors that accidentally recalculate the results _every time_.  This is primarily due to badly-written _input functions_, such as:

```ts
createSelector(
  state => ({ foo: state.foo, bar: state.bar }), 
  ({ foo, bar }) => ({result: foo + bar})
)
```

Since the input function _always_ returns a new object, the output function will _always_ run and return a new reference.

Probably the simplest thing we can do here would be to double-run the input functions twice and verify that the results are shallow equal.

EskiMojo14:
fixes #611


src/index.ts
```diff
@@ -111,6 +115,9 @@ export function createSelectorCreator<
 
     const dependencies = getDependencies(funcs)
 
+    // do we want to try and use the equality fn from options? how do we account for other memoizers?
+    const cacheKeyComparator = createCacheKeyComparator(defaultEqualityCheck)
```
markerikson, line 119: Straight shallow equality is probably simplest, although I _can_ see an argument for using the specific memoization setup they provided.

src/index.ts
```diff
@@ -133,10 +140,32 @@ export function createSelectorCreator<
...
       const length = dependencies.length
 
+      const paramsCopy = []
+
       for (let i = 0; i < length; i++) {
         // apply arguments instead of spreading and mutate a local list of params for performance.
         // @ts-ignore
         params.push(dependencies[i].apply(null, arguments))
+
+        if (process.env.NODE_ENV !== 'production' && inputStabilityCheck) {
```
markerikson, line 150: :hand: Can we move _all_ the logic inside the main `NODE_ENV` check for simplicity?  I'm fine with copy-pasting the loop.

EskiMojo14: sure, i just figured two loops would be worse for performance than one

EskiMojo14: tests say i was wrong 🎉

src/index.ts
```diff
@@ -38,6 +38,14 @@ export { defaultMemoize, defaultEqualityCheck }
 
 export type { DefaultMemoizeOptions }
 
+type StabilityCheck = boolean | 'once'
```
markerikson, line 41: :hand: How about we just change this to an enum?

`'always' | 'once' | 'never'`

src/index.ts
```diff
@@ -120,6 +127,15 @@ export function createSelectorCreator<
       ...finalMemoizeOptions
     )
 
+    // @ts-ignore
+    const makeAnObject: (...args: unknown[]) => object = memoize(
```
markerikson, line 131: :hand: Should this be moved inside the `NODE_ENV` check?

src/index.ts
```diff
@@ -131,16 +147,49 @@ export function createSelectorCreator<
...
+          // @ts-ignore
+          paramsCopy.push(dependencies[i].apply(null, arguments))
+        }
+
+        // if the memoize method thinks the parameters are equal, these *should* be the same reference
+        const equal =
+          makeAnObject.apply(null, params) ===
+          makeAnObject.apply(null, paramsCopy)
+        if (!equal) {
+          // do we want to log more information about the selector?
```
markerikson, line 178: :speech_balloon: ~~What happens if we reference `selector.name` here?  Does it end up as `"selector"`, or whatever we assigned it to?~~

Scratch that, it comes back as `"memoized"`, because `defaultMemoize.ts` has `function memoized()`.

Mmm... We could at least log out the original inputs and both sets of extracted values?

test/inputStabilityCheck.spec.ts
```diff
@@ -0,0 +1,112 @@
+import { createSelector, setInputStabilityCheckEnabled } from 'reselect'
+import { shallowEqual } from 'react-redux'
```
markerikson, line 2: :Hand: strictly speaking we could do this with the `createCacheKeyComparator()` util.  And for that matter, given that we've got it being called in one or two places, maybe we ought to just extract `const shallowEqual = createCacheKeyComparator(defaultEqualityCheck)` into a `utils.ts` file and reuse that.

markerikson (commented):
Looks good overall.  Can you make these couple tweaks?

src/index.ts
EskiMojo14: it could, but i think that would be worse for performance? this creates one function when the selector is created, moving it inside would make a new function every time the check runs

test/inputStabilityCheck.spec.ts
EskiMojo14: createCacheKeyComparator assumes an array, which { a, b } is not 🤔

markerikson: Ah, so it does. Feels annoying pulling in React-Redux just for a shallow equality function, but I guess it's dev-only.

src/index.ts
EskiMojo14: added 
```ts
{
  arguments,
  firstInputs: params,
  secondInputs: paramsCopy
}
```
open to better names 🙂

test/inputStabilityCheck.spec.ts
EskiMojo14: yeah, that was my rationale - it's just for tests, doesn't affect the library user

PR #622, opened 2023-08-21
Record computation time

thaihuyhung:
Hi,

Currently, I'm using `recomputations` count to verify the performance of our selectors in local environment. This gave us a lot of insights into how we implemented the selectors. Besides the re-computation count, I'd like to know more about how long is the execution/computation time as well. Therefore I create this PR for this change. Hope it would be helpful for everyone.

Sorry if I didn’t follow any guidelines, this is the first time I submit PR.

Thank you.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -46,6 +46,14 @@ export function setInputStabilityCheckEnabled(enabled: StabilityCheck) {
   globalStabilityCheck = enabled
 }
 
+function getTime() {
+  if (performance && typeof performance.now === 'function') {
+    return performance.now()
+  }
+
+  return Date.now()
+}
+
 function getDependencies(funcs: unknown[]) {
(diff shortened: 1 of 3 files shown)


aryaemami59:
Looks very interesting, can we disable during production mode?

src/index.ts
```diff
@@ -46,6 +46,14 @@ export function setInputStabilityCheckEnabled(enabled: StabilityCheck) {
   globalStabilityCheck = enabled
 }
 
+function getTime() {
```
appmateeu, line 49: I am wondering if wouldn't be better if definition of `getTime` would look like:
```typescript
const getTime = performance && typeof performance.now === 'function'
  ? performance.now.bind(performance)
  : Date.now.bind(Date)
```

In that case no extra conditions or jumps will be done while the function executes.

PR #623, opened 2023-09-13
Use default Node version in CI

nickmccurdy:
Deprecated Node versions are used in CI, potentially causing security and reliability issues. Instead, it's better to use GitHub's default Node version, which also doesn't require additional downloads or installations.


nickmccurdy:
Blocked by https://github.com/reduxjs/redux-toolkit/pull/3715

PR #625, opened 2023-09-14
Typed createSelector function-solution-2

aryaemami59:
This is the second solution for #624. It is not as straightforward but it is done without modifying any of the original types.

```
export const defaultCreateSelector = createSelectorCreator(defaultMemoize)
export const weakMapCreateSelector = createSelectorCreator(weakMapMemoize)
export const AutotrackCreateSelector = createSelectorCreator(autotrackMemoize)

export type TypedCreateSelectorDefault<State> = <
  Selectors extends readonly Selector<State>[],
  Result
>(
  ...args: Parameters<typeof defaultCreateSelector<Selectors, Result>>
) => ReturnType<typeof defaultCreateSelector<Selectors, Result>>

export type TypedCreateSelectorWeakMap<State> = <
  Selectors extends readonly Selector<State>[],
  Result
>(
  ...args: Parameters<typeof weakMapCreateSelector<Selectors, Result>>
) => ReturnType<typeof weakMapCreateSelector<Selectors, Result>>
// Not sure if this is necessary since the options object for both weakMapMemoize and autotrackMemoize is identical.
export type TypedCreateSelectorAutotrack<State> = <
  Selectors extends readonly Selector<State>[],
  Result
>(
  ...args: Parameters<typeof AutotrackCreateSelector<Selectors, Result>>
) => ReturnType<typeof AutotrackCreateSelector<Selectors, Result>>
```

I would love to help and contribute as much as I can. So please let me know if there is anything else I can do as I would be more than happy to help.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,40 +1,42 @@
 import type {
-  Selector,
+  DropFirst,
+  Expand,
   GetParamsFromSelectors,
+  Head,
+  MergeParameters,
+  ObjValueTuple,
   OutputSelector,
+  Selector,
   SelectorArray,
   SelectorResultArray,
-  DropFirst,
-  MergeParameters,
-  Expand,
-  ObjValueTuple,
-  Head,
   Tail
 } from './types'
 
 export type {
-  Selector,
+  EqualityFn,
   GetParamsFromSelectors,
   GetStateFromSelectors,
-  OutputSelector,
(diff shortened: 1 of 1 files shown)


aryaemami59:
See https://github.com/reduxjs/reselect/pull/624#issuecomment-1783931536

PR #633, opened 2023-10-31
Fix hover preview of generated selectors.

aryaemami59:
This PR aims to flatten and expand the types for generated selectors in order to improve their visual display inside hover previews.

<details>

  <summary>

## This is what it looked like before the changes were made:

</summary>

Inside `typescript_test/test.ts` on line 1269:

before

</details>

<details>

  <summary>

## And this is what it looks like after changes were made:

</summary>

Same selector inside `typescript_test/test.ts` now on line 1319:

after

</details>

Other minor changes include:

- [x] Remove left over types related to `ts46-MergeParameters`.
- [x] Reorganized `types.ts` by category.
- [x] Fix `createStructuredSelector` type inference.
- [x] Remove second overload of `StructuredSelectorCreator`, since the function itself has no behavior indicating it would need multiple function signatures.
- [x] Create `TypedStructuredSelectorCreator` which is used to effectively replace `StructuredSelectorCreator` second function signature (WIP).
- [x] Add more JSDocs.
- [x] Add `testUtils.ts` which contains some utilities that will aid in unit testing.
- [x] Add more type tests.
- [x] Add more unit tests.
- [x] Fix infinite type instantiation issue that was caused inside of `deepNesting` function in `typescript_test\test.ts` file.
- [x] Add public and internal tags to existing JSDocs.
- [x] Remove non-existent rules from `.eslintrc`.
- [x] Add `@typescript-eslint/consistent-type-imports` rule to help with auto-fixing type imports.


aryaemami59:
@markerikson This also fixes the TS recursion problem so now we can do this:

deep-nesting

and this:

param-list

markerikson:
ho. ly. smoke.

that is _amazing_!

lemme actually read through the changes, but wow, I'm seriously impressed at the apparent results!

aryaemami59:
> ho. ly. smoke.
> 
> that is _amazing_!
> 
> lemme actually read through the changes, but wow, I'm seriously impressed at the apparent results!

Thank you, it feels good to hear that. I don't know if I made it clear enough in the JSDocs and commit messages, but the reason the infnite instantiation bug was happening is because, TypeScript was passing generic parameters into other types that then took those generic parameters and tried to instantiate a type by doing some calculation. The problem was that it was trying to do all of the instantiation using recursion all at the same time which caused the bug in the first place. What I did was that I tried to interrupt the recursion process by expanding and flattening the types (in this case `OutputSelector`) right before the recursion process begins. So TypeScript now instantiates the type and caches it before recursion so it doesn't have to do all of the work at the same time. I might be completely wrong, but to me it seems like that is what is happening.

markerikson:
Okay, skimmed through. Obviously that's a pretty big diff, enough so that we're definitely into "LGTM 🤷‍♂️ !" territory :)  But the changes that I saw seemed to make sense, and that hover output is _exactly_ what I was hoping to see.  Honestly the only tiny half-nitpick I had was that the JSDoc example for `createStructuredSelector` probably doesn't need to have the actual items arrays filled out, just to save some visual space.

But no reason to hold up this PR for that :)

Meanwhile, totally unrelated to _this_ repo: we're having some active debates about the design of the new `createSlice`.selectors` field. Latest set of questions is in https://github.com/reduxjs/redux-toolkit/discussions/3387#discussioncomment-7429635 .  Would love to have you offer some feedback and help us figure that out so we can finish RTK 2.0!

aryaemami59:
> Okay, skimmed through. Obviously that's a pretty big diff, enough so that we're definitely into "LGTM 🤷‍♂️ !" territory :) But the changes that I saw seemed to make sense, and that hover output is _exactly_ what I was hoping to see. Honestly the only tiny half-nitpick I had was that the JSDoc example for `createStructuredSelector` probably doesn't need to have the actual items arrays filled out, just to save some visual space.
> 
> But no reason to hold up this PR for that :)
> 
> Meanwhile, totally unrelated to _this_ repo: we're having some active debates about the design of the new `createSlice`.selectors` field. Latest set of questions is in [reduxjs/redux-toolkit#3387 (comment)](https://github.com/reduxjs/redux-toolkit/discussions/3387#discussioncomment-7429635) . Would love to have you offer some feedback and help us figure that out so we can finish RTK 2.0!

You're right about the JSDocs, my apologies, I will fix that right away. About the `createSlice.selectors`, I was actually working with them today, I will join the discussion shortly. And please let me know if there anything else I can help with.

PR #628, opened 2023-10-20
fix GitHub workflow badge URL

shoito:
fix GitHub workflow badge URL.

See https://github.com/badges/shields/issues/8671

> You need to update your badge URL from
> https://img.shields.io/github/workflow/status/<user>/<repo>/Run%20Tests
> to
> https://img.shields.io/github/actions/workflow/status/<user>/<repo>/test.yml?branch=main

PR #621, opened 2023-08-14
docs: add instructions for bun and pnpm

colinhacks:

Issue #598, opened 2023-02-11
Contradiction between usage examples in README and Redux docs


DenyVeyten:
Basic Usage examples https://github.com/reduxjs/reselect#basic-usage (`selectSubtotal`, `selectTax`) contradict guide in Redux docs https://redux.js.org/usage/deriving-data-selectors#balance-selector-usage

I think both should be updated to some middleground, explaining that memoization is good over heavy operations (e.g. iterating long lists), but is overhead for simple ones (e.g. small math calculations, iterating short lists).

markerikson:
Yeah, the README here is rather dated, but we've got a lot of other priorities and this really isn't on the list to update any time soon.

aryaemami59:
This one has been resolved by #636.

Issue #515, opened 2021-10-20
Rewrite the readme to show "Modern Redux" patterns and practices


markerikson:
Per #355 , the README now at least uses `selectX` as the naming convention, but all the examples are still based on `connect` and "hand-written legacy Redux" patterns.

We need to rewrite those to show use of function components, Redux Toolkit, and `useSelector` instead, as well as the advice in the new [Deriving Data with Selectors](https://redux.js.org/usage/deriving-data-selectors) docs page.

PR #636, opened 2023-11-11
Update Docs for v5 changes.

aryaemami59:
This PR mostly focuses on updating the documentation for v5.

This includes removing any examples that use legacy syntax such as `connect` or `switch` statements.
Currently there are some confusing jargon related to Reselect, things like "output selector" and "result function" are sometimes used interchangeably which can cause more confusion down the line. To mitigate that problem I also added a "Terminology" section. There were a lot of changes made to the documentation so far but there will be some more coming. I will probably make some more examples with CodeSandbox to demonstrate how the new features work.

 - [x] Add documentation regarding new features.
- [x] Redo CodeSandbox examples to align better with documentation. Solves https://github.com/reduxjs/reselect/issues/598.
- [x] Fix GitHub workflow badge URL https://github.com/reduxjs/reselect/pull/628.
- [x] Add instructions for `bun` and `pnpm`. Solves https://github.com/reduxjs/reselect/pull/621.
- [x] `argsMemoize` and `argsMemoizeOptions` solve https://github.com/reduxjs/reselect/pull/359.
- [x] Remove Redux legacy patterns including `connect` and `switch` statements from docs.  Solves https://github.com/reduxjs/reselect/issues/515.
- [x] Replace legacy code patterns with modern syntax like `useSelector` and functional components.
- [x] Implementation of `argsMemoize` solves https://github.com/reduxjs/reselect/issues/376.
- [x] Document order of execution in Reselect. Solves https://github.com/reduxjs/reselect/issues/455.


README.md
```diff
@@ -1,707 +1,1799 @@
 # Reselect
 
+[![npm package][npm-badge]][npm][![Coveralls][coveralls-badge]][coveralls][![GitHub Workflow Status][build-badge]][build]![TypeScript][typescript-badge]
+
 A library for creating memoized "selector" functions. Commonly used with Redux, but usable with any plain JS immutable data as well.
 
-- Selectors can compute derived data, allowing Redux to store the minimal possible state.
-- Selectors are efficient. A selector is not recomputed unless one of its arguments changes.
-- Selectors are composable. They can be used as input to other selectors.
+- Selectors can compute derived data, allowing `Redux` to store the minimal possible state.
```
markerikson, line 7: :hand: style note: we typically leave the name of the library as a proper capitalized name ("Reselect", "Redux", "React-Redux" "Redux Toolkit"), and specifically use backticks for the lower-case actual package name (`reselect`, `redux`, `react-redux`, `@reduxjs/toolkit`).  Could you fix the names?

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+```ts
+import { createSelector } from '@reduxjs/toolkit'
+```
+
+### Standalone
+
+For standalone usage, install the `Reselect` package:
+
+#### Using `npm`
```
markerikson, line 27: :hand: Two notes:

- Not sure we to list all four package managers separately
- However many we leave in, consolidate them into a single code block ( `bash` language, and a `# NAME` comment above, like:

```bash
# NPM
npm install reselect

# Yarn
yarn add reselect
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+```bash
+pnpm add reselect
+```
+
+---
+
+## Basic Usage
+
+`Reselect` exports a [`createSelector`] API, which generates memoized selector functions. [`createSelector`] accepts one or more [`input selectors`], which extract values from arguments, and a [`combiner`] function that receives the extracted values and should return a derived value. If the generated [`output selector`] is called multiple times, the output will only be recalculated when the extracted values have changed.
```
markerikson, line 55: :hand: 

- There's a bunch of inline code names that also have brackets.  Are these supposed to be links?
- Typically inline code formatting is used for actual variable / function names.  "combiner" and "input selector" are descriptive terms and shouldn't be formatted as code.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+console.log(selectCompletedTodos(state) === selectCompletedTodos(state)) //=> false
+
+console.log(
+  memoizedSelectCompletedTodos(state) === memoizedSelectCompletedTodos(state)
+) //=> true
+```
+
+As you can see from the example above, `memoizedSelectCompletedTodos` does not run the second or third time, but we still get the same return value as last time.
+
+Another difference is that with `memoizedSelectCompletedTodos` the referential integrity of the return value is also maintained through multiple calls of the selector, but the same cannot be said about the first example.
```
markerikson, line 108: :hand: the phrasing of this sentence feels a bit awkward.  How about:

```suggestion
In addition to skipping unnecessary recalculations, `memoizedSelectCompletedTodos` returns the existing result reference if there is no recalculation.  This is important for libraries like React-Redux or React that often rely on reference equality checks to optimize UI updates.
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+- [License](#license)
+- [Prior Art and Inspiration](#prior-art-and-inspiration)
+
+---
+
+## Terminology
+
+<details><summary><b>Click to expand</b></summary>
+
+- <a name="selector-function"></a>[**`Selector Function`**](#selector-function): A function that takes the `Redux` store state (or a part of it) as an argument and returns data derived from that state.
```
markerikson, line 158: :hand: I think it's worth keeping the "not actually required to use Redux" aspect here.  How about:

```suggestion
- <a name="selector-function"></a>[**`Selector Function`**](#selector-function): A function that accepts one or more JS values as input arguments, and derives a result.  When used with Redux, the first argument is typically the entire Redux store state.
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+- <a name="input-selectors"></a>[**`input selectors`**](#input-selectors): Basic selector functions used as building blocks for creating a memoized selector. They are passed as the first argument(s) to [`createSelector`].
```
markerikson, line 159:
```suggestion
- <a name="input-selectors"></a>[**`input selectors`**](#input-selectors): Basic selector functions used as building blocks for creating a memoized selector. They are passed as the first argument(s) to [`createSelector`], are called with all of the selector arguments, and extract values for use in the result function.
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+- <a name="output-selector"></a>[**`Output Selector`**](#output-selector): The actual memoized selectors created by [`createSelector`].
+- <a name="result-function"></a>[**`Result Function (Combiner)`**](#result-function): The function that comes after the [`input selectors`]. It takes the [`input selectors`]' return values as arguments and returns a result. Otherwise known as the `combiner`.
+- <a name="combiner"></a>[**`Combiner`**](#combiner): A function that takes [`input selectors`]' return values as arguments and returns a result (This function comes after the [`input selectors`] inside [`createSelector`]). This term is somewhat interchangeably used with `Result Function` or `resultFunc`. But `combiner` is more of a general term and `resultFunc` is more specific to `Reselect`. So the `resultFunc` is a `combiner` but a `combiner` is not necessarily the same as `resultFunc`. For the sake of simplicity, they are used synonymously throughout this documentation.
```
markerikson, line 162: :hand: if we're going to consolidate terms, I'd like to drop "combiner" entirely and just stick with "result function" (or possibly "output function").

markerikson (changes requested):
Only got through the first few sections so far - I'll come back and review the rest later.  Wanted to pass on the initial feedback.  Thanks for working on this!

README.md
aryaemami59: I only did that to make copy pasting with a single click easier. Do you still want me to change it?

README.md
aryaemami59: - Yes the brackets are links.
- I will fix the inline formatting.

Thanks for pointing these out!

README.md
aryaemami59: Sounds good, the only reason I included "combiner" is because of how it was referened internally. But you're right we can drop that term in the docs to avoid further confusion. I think we can just stick to "result function". "output function" can be confused with "output selector". What do you think?

README.md
aryaemami59: Yes sir, sorry about that. I will fix that right away!

README.md
markerikson: Oh. I missed the use of "output selector" here.  I don't think I like that usage tbh :)  I personally have used that phrase to mean what we're now calling the "result function".  If we have to distinguish things, I'd say "generated selector", maybe?

README.md
markerikson: I don't think we need to optimize for people copy-pasting the instructions to install one package, yeah :)  We also use the code block approach in the other Redux docs sites for this step anyway.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+---
+
+## How Does `Reselect` Work?
+
+<a id="cascadingmemoization"></a>
+
+### Cascading Memoization
+
+<details><summary><b>Click to expand</b></summary>
```
markerikson, line 186: :hand: couple thoughts here:

I'd recently added this snippet under the "how do you pass in arguments?" section:

```ts
const finalSelector = (...args) => {
  const extractedValues = inputFunctions.map(input => input(...args));
  return output(...extractedValues);
}
```

I'd like to have that right under the `## How Does Reselect Work?` header, and one or two short paragraphs of explanation.  

Also, I'd like to avoid having headers with _no_ content at all. Let's have at least a sentence or two under each header with a basic explanation, then use the `<details>` to let them dig in further.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+a single ["result function"][`result function`] / ["combiner"][`combiner`], and an optional options object, and
+generates a memoized selector function.
+
+</details>
+
+<details><summary><b>Parameters</b></summary>
+
+| Name                     | Description                                                                                      |
+| :----------------------- | :----------------------------------------------------------------------------------------------- |
+| `inputSelectors`         | An array of [`input selectors`], can also be passed as separate arguments.                       |
```
markerikson, line 311: :hand: We'll to update the terms consistently, but especially here.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+---
+
+## API
+
+<a id="createselector"></a>
+
+### createSelector(...inputSelectors | [inputSelectors], resultFunc, createSelectorOptions?)
+
+<details><summary><b>Description</b></summary>
```
markerikson, line 299: :hand: "Description", "Parameters", and "Returns" shouldn't be hidden inside of `<details>` - we want readers to be able to see those directly as they scroll down.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+<a id="defaultmemoize"></a>
+
+#### defaultMemoize(func, equalityCheckOrOptions = defaultEqualityCheck)
+
+<details><summary><b>Description</b></summary>
+
+The standard memoize function used by [`createSelector`].
+
+It has a default cache size of 1. This means it always recalculates when the value of an argument changes. However, this can be customized as needed with a specific max cache size (**`Since`** 4.1.0).
```
markerikson, line 348: Don't need to format "since" anywhere:

```suggestion
It has a default cache size of 1. This means it always recalculates when the value of an argument changes. However, this can be customized as needed with a specific max cache size (since 4.1.0).
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+</details>
+
+<details><summary><b>Parameters</b></summary>
+
+| Name                     | Description                                                 |
+| :----------------------- | :---------------------------------------------------------- |
+| `func`                   | The function to be memoized.                                |
+| `equalityCheckOrOptions` | Either an `equality check` function or an `options` object. |
+
+**`Since`** 4.1.0, `defaultMemoize` also accepts an options object as its first argument instead of an `equalityCheck` function. The `options` object may contain:
```
markerikson, line 367:
```suggestion
Since 4.1.0, `defaultMemoize` also accepts an options object as its first argument instead of an `equalityCheck` function. The `options` object may contain:
```

README.md
aryaemami59: Honestly I think "output selector" makes sense for the generated selector. It would also be consistent with how the types were named initially. The result function is not really a "selector" if you think about it. The way it is now we have "input selectors", then "result function" does the calculation, and we get an "output selector". What do you think?

README.md
```diff
@@ -1,707 +1,1799 @@
...
+| `resultEqualityCheck` | If provided, used to compare a newly generated output value against previous values in the cache. If a match is found, the old value is returned. This addresses the common <code>todos.map(todo => todo.id)</code> use case, where an update to another field in the original data causes a recalculation due to changed references, but the output is still effectively the same. |
+| `maxSize`             | The cache size for the selector. If greater than 1, the selector will use an LRU cache internally. <br /> **`Default`** = 1                                                                                                                                                                                                                                                         |
+
+> [!WARNING]
```
markerikson, line 385: :hand: This "warning" block doesn't seem to be styled right - should be this, I think:

```suggestion
> **Warning**
```

README.md
aryaemami59: So it's kind of like:

input -> calculation -> output

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+<details><summary><b>Type parameters</b></summary>
+
+| Name   | Description                                |
+| :----- | :----------------------------------------- |
+| `Func` | The type of the function that is memoized. |
+
+</details>
+
+<details><summary><b><code>Examples</code></b></summary>
```
markerikson, line 404: :hand: don't need to format "Examples":

```suggestion
<details><summary><b>Examples</b></summary>
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+###### Using [`createSelector`]
```
markerikson, line 406: :hand: Can we tweak these two titles?  Maybe emphasize "Using `defaultMemoize` with `createSelector`", etc.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+</details>
+
+<div align="right">[ <a href="#table-of-contents">↑ Back to top ↑</a> ]</div>
+
+---
+
+<a id="autotrackmemoize"></a>
+
+#### autotrackMemoize(func) - (**`Since`** 5.0.0)
```
markerikson, line 662: :hand: Let's 100% refer to this as `unstable_autotrackMemoize` everywhere in the docs until it's actually ready for release.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+<details><summary><b>Description</b></summary>
+
+Uses an "auto-tracking" approach inspired by the work of the Ember Glimmer team. It uses a Proxy to wrap arguments and track accesses to nested fields in your selector on first read. Later, when the selector is called with new arguments, it identifies which accessed fields have changed and only recalculates the result if one or more of those accessed fields have changed. This allows it to be more precise than the shallow equality checks in defaultMemoize.
```
markerikson, line 666: :hand: Add a warning under this section clarifying that it is still experimental and undergoing testing

README.md
```diff
@@ -1,707 +1,1799 @@
...
+  memoize: autotrackMemoize
+})
+
+const selectTodoIds = createSelectorAutotrack(
+  [(state: RootState) => state.todos],
+  todos => todos.map(todo => todo.id)
+)
+```
+
+- Pros:
```
markerikson, line 753: :hand: Looks like there's some copy-paste duplication from the description above? Can remove this.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+  (todos, id) => todos.filter(todo => todo.id === id)
+)((state: RootState, id: number) => id)
+```
+
+Starting in 5.0.0, you can eliminate this problem using `weakMapMemoize`.
+
+###### Using [`createSelector`]
+
+```ts
+const parametricSelector = createSelector(
```
markerikson, line 616: :hand: tbh I've never liked the phrase "parametric selector", and I don't think using it as a name in this example helps.  

Also, doing a `.filter()` with _one_ ID in comparison seems unlikely :)  How about filtering by "category" instead?  Seems like something that you might want to do many times.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+| Name   | Description                                |
+| :----- | :----------------------------------------- |
+| `Func` | The type of the function that is memoized. |
+
+</details>
+
+<details><summary><b><code>Examples</code></b></summary>
+
+Prior to `weakMapMemoize`, you had this problem:
```
markerikson, line 522: :hand: This section feels like a "use case comparison", rather than "examples".

README.md
```diff
@@ -1,707 +1,1799 @@
...
+```ts
+import { createCachedSelector } from 're-reselect'
+
+const parametricSelector = createCachedSelector(
+  [(state: RootState) => state.todos, (state: RootState, id: number) => id],
+  (todos, id) => todos.filter(todo => todo.id === id)
+)((state: RootState, id: number) => id)
+```
+
+Starting in 5.0.0, you can eliminate this problem using `weakMapMemoize`.
```
markerikson, line 611: :hand: If we're going to have this kind of comparison, there ought to be more information here explaining _why/how_ `weakMapMemoize` solves this issue.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+</details> -->
+
+<div align="right">[ <a href="#table-of-contents">↑ Back to top ↑</a> ]</div>
+
+---
+
+<a id="createselectorcreator" ></a>
+
+### createSelectorCreator(memoize | options, ...memoizeOptions)
```
markerikson, line 779: :hand: let's move this up to be directly after `createSelector`.  I think the reasonable ordering is:

- `createSelector`
- `createSelectorCreator`
- `createStructuredSelector`
- memoization functions

README.md
```diff
@@ -1,707 +1,1799 @@
...
+---
+
+<a id="createstructuredselector"></a>
+
+### createStructuredSelector({ inputSelectors }, selectorCreator = createSelector)
+
+<details><summary><b>Description</b></summary>
+
+A convenience function for a common pattern that arises when using `Reselect`.
+The selector passed to a `connect` decorator often just takes the values of its [`input selectors`]
```
markerikson, line 926: :hand: let's ditch the `connect` reference, and instead say it simplifies returning an object made up of selector results.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+```
+
+| Possible Values | Description                                     |
+| :-------------- | :---------------------------------------------- |
+| `once`          | Run only the first time the selector is called. |
+| `always`        | Run every time the selector is called.          |
+| `never`         | Never run the `input stability check`.          |
+
+> [!IMPORTANT]
+> The `input stability check` is automatically disabled in production environments.
```
markerikson, line 1066: :hand: don't need to format words like these.

```suggestion
> The input stability check is automatically disabled in production environments.
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+| Name                            | Description                                                                                                                                                                                       |
+| ------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `resultFunc`                    | The final function passed to [`createSelector`]. Otherwise known as the `combiner`.                                                                                                               |
+| `memoizedResultFunc`            | The memoized version of `resultFunc`.                                                                                                                                                             |
+| `lastResult`                    | Returns The last result calculated by `memoizedResultFunc`.                                                                                                                                       |
```
markerikson, line 1120:
```suggestion
| `lastResult`                    | Returns the last result calculated by `memoizedResultFunc`.                                                                                                                                       |
```

README.md
```diff
@@ -1,707 +1,1799 @@
...
+<a id="v5summary"></a>
+
+## What's New in 5.0.0?
+
+<details><summary><b>Version 5.0.0 introduces several new features and improvements:</b></summary>
+
+- **Customization Enhancements**:
+
+  - Added the ability to pass an options object to [`createSelectorCreator`], allowing for customized `memoize` and `argsMemoize` functions, alongside their respective options (`memoizeOptions` and `argsMemoizeOptions`).
+  - The [`createSelector`] function now supports direct customization of `memoize` and `argsMemoize` within its options object.
```
markerikson, line 1144: :hand: I'd put this one above the `createSelectorCreator` customization, really :)

src/createCurriedSelectorCreator.ts
markerikson, line 1: :hand: I'd like to drop `createCurriedSelector` from this PR entirely, for a few reasons:

- This PR is huge already, and I'd like it to focus on just the docs changes
- Not yet sure it's something we actually want to ship
- It's something we can discuss after 5.0 is done

src/types.ts
markerikson, line 1: :hand: Can you also move all the types changes to a separate PR as well? That would make this easier to review and keep track of

test/benchmarks/orderOfExecution.bench.ts
markerikson, line 1: :hand: Heh, yep, also please move new benchmarks to a separate PR as well :)

Can probably split this into at least 4 PRs:

- README docs changes
- Types changes
- Benchmarks
- `createCurriedSelector`

markerikson (changes requested):
Okay, got some additional requests:

- We can limit terminology links to just the first time each term gets used, as well as maybe the first usage for each inside the API reference section
- Got some requests for reordering the API reference
- Several terms that are bolded or inline-code that don't need to be ("Since", "Example", etc)
- Can you split this into multiple PRs? this is too much to review at once :)  and they're really at least 4 separate topics:
  - Docs changes
  - Types changes
  - Benchmarks
  - `createCurriedSelector`

README.md
markerikson: eh, I'm also not sold on the type names either, even though they've been there since early on :)  "Parametric selector" has always confused me.

I think I'd like to stick with either "generated selector" or "memoized selector" as the term for the function returned by `createSelector`.

aryaemami59: > eh, I'm also not sold on the type names either, even though they've been there since early on :) "Parametric selector" has always confused me

Tbh, me neither, but I was just trying to stay consistent with the type names. And you're right about the "parametric selector", I'm not really a fan of it either. But they were there since the beginning and I thought that's how you would want me to name things in the docs as well.

aryaemami59: > I think I'd like to stick with either "generated selector" or "memoized selector" as the term for the function returned by `createSelector`.

"memoized selector" makes sense. However, since we have "input selectors" and if we have an "input" it makes sense that we would have an "output" as well. So I still think "output selector" makes the most sense.

aryaemami59: Also we have 2 types that I would like to remove entirely if that's okay with you. `ParametricSelector` and `OutputParametricSelector` since `Selector` and `OutputSelector` support additional parameters.

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+     - If they differ, it proceeds to the second level.
+
+   - **Second Level**: It runs the [`input selectors`] and compares their current results with the previous ones (done by `memoize`).
+     > [!NOTE]
+     > If any one of the [`input selectors`] return a different result, all [`input selectors`] will recalculate.
+     - If the results are the same, it returns the cached result without running the [`result function`].
+     - If the results differ, it runs the [`result function`].
+
+This behavior is what we call **_Cascading Double-Layer Memoization_**.
```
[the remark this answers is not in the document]
aryaemami59: If `argsMemoize` fails, Reselect "cascades" to `memoize`. I will tweak the explanation to make it more clear.

README.md
aryaemami59: > I'd recently added this snippet under the "how do you pass in arguments?" section:

Honestly the snippet confused me, but I think I know what you're trying to convey. And I already had an explanation of what you're trying to get across with the snippet, this is what it looks like:

```ts
const selectAvailableItems = createSelector(
  [
    // First input selector extracts items from the state
    (state: RootState) => state.items,
    // Second input selector forwards the category argument
    (state: RootState, category: string) => category,
    // Third input selector forwards the ID argument
    (state: RootState, category: string, id: number) => id
  ],
  // Output selector uses the extracted items, category, and ID
  (items, category, id) =>
    items.filter(item => item.category === category && item.id !== id)
)
```

Internally Reselect is doing this:

```ts
// Input selector #1
const items = (state: RootState, category: string, id: number) => state.items
// Input selector #2
const category = (state: RootState, category: string, id: number) => category
// Input selector #3
const id = (state: RootState, category: string, id: number) => id
// result of `output selector`
const finalResult =
  // The `Result Function`
  items.filter(item => item.category === category && item.id !== id)
```

> Also, I'd like to avoid having headers with no content at all. Let's have at least a sentence or two under each header with a basic explanation, then use the <details> to let them dig in further.

That makes a lot of sense, I will fix it. :)

README.md
aryaemami59: What "term" are we talking about here?

README.md
aryaemami59: Yessirrr. I will adjust that.

README.md
aryaemami59: Initially got that from [markdown-magic](#https://github.com/DavidWells/markdown-magic). But you're right I will fix it :)

README.md
aryaemami59: They're the new [alert badges](#https://github.com/orgs/community/discussions/16925) in github-flavored markdown

README.md
aryaemami59: yesssirrr

README.md
markerikson: Yeah, go ahead and remove those. Breaking change, we'll have to document that, but might as well.

Per the "input/output" thing: I think I disagree because the "input selectors" are "selectors" (args in, derived value out even if it's usually just an extraction), and "inputs to `createSelector`".  The "result function" doesn't get any of the original args, so even though I've _referred_ to it as an "output selector" in the past, that's really not a good usage and I shouldn't have called it that.

The actual generated selector isn't just about "output", though.  It has all three pieces: the input arguments, the memoization, and calculating the final result.  It's not _just_ about "output".  The overall signature is `(...args) => derivedValue`, so it's a standard selector overall, it just happens to be a "memoized" selector that was "generated" by `createSelector` rather than hand-written.  So, I do think that "memoized" or "generated" are the best terms here.

README.md
aryaemami59: Surething

README.md
markerikson: Yeah, my example is meant to show that "_all_ arguments get passed to _all_ input selectors", specifically (which is why they all have to have compatible signatures).

README.md
aryaemami59: I will update the example, do you want me to name it something generic, what would we call a selector that takes additional parameters? or perhaps we don't actually need to specify that and we can call it something generic "selectSomething"? to stay consistent with the "selectX" pattern?

README.md
markerikson: Good question, what _was_ I thinking? :)

It was one of two things:

- We can skip wrapping every single use of one of the "Terminology" terms in links.  Per one of my other comments, we only need to highlight those in the first usage (and we only want to use backticks for actual variable/function names)
- Making sure that whatever final terminology terms we settle on get used consistently, particularly in the API reference section

README.md
aryaemami59: I will the additional info. Just to be sure, we are specifically refering to the "cache size" problem right?

README.md
markerikson: Huh, okay, hadn't seen that variation.

README.md
aryaemami59: It was ordered that way initially but you're correct, I will rearrange them.

README.md
markerikson: Try to name them meaningfully based on the example (like `selectTodosByCategory`, etc), or just `selectSomething` as a literal fallback, yeah

README.md
aryaemami59: I could not agree more. It seems like `createStructuredSelector` was created specifically to be used with `connect` and now that `connect` is legacy API, it seems like `createStructuredSelector` is also kind of a legacy API.

README.md
aryaemami59: Yes sire!

README.md
aryaemami59: You have the eyes of an eagle!

README.md
aryaemami59: You wish is my command :)

README.md
aryaemami59: ok :(

README.md
aryaemami59: Yeah the "result function" is not an "output selector". I agree. That was one of the things that confused me when I first started working on the codebase. the "result function" was sometimes referred to as "the combiner" (which I think in terms of semantics makes the most sense). But it was occasionally referred to as the "output selector", but the genereated selector was also sometimes referred to as the "output selector". In the JSDocs I always called the final generated selector "a memoized output selector". 

BTW I did a quick search in github on `ParametricSelector` and `OutputParametricSelector` seems like anybody was hardly using them so I think we'll be okay removing them, I will go ahead and document their removal.

I still think in terms of simplicity "output selector" makes sense. The "memoized" part is implied and kind of a given. Input -> result function -> output
 sums up the flow of code as well. It's self explanatory in that Reselect goes through the "inputs" then "resultFunc" and gives an "output". It's linear and easier to digest no?

README.md
```diff
@@ -1,707 +1,1799 @@
...
+
+<a id="cascadingmemoization"></a>
+
+### Cascading Memoization
+
+<details><summary><b>Click to expand</b></summary>
+
+The way `Reselect` works can be broken down into multiple parts:
+
+1. **Initial Run**: On the first call, `Reselect` runs all the [`input selectors`], gathers their results, and passes them to the [`result function`].
```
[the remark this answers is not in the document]
aryaemami59: Just wanted to make jumping to definitions easier, kind of like how VSCode works with TS and type definitions.

README.md
aryaemami59: This is how I explained it:

1. **Consistency in Arguments**: Ensure that all positional arguments across [input selectors] are of the same type for consistency.

2. **Selective Argument Usage**: Design each selector to use only its relevant argument(s) and ignore the rest. This is crucial because all [input selectors] receive the same arguments that are passed to the [output selector].

Do you still want me to include your snippet as well?

README.md
aryaemami59: Yes sir, I will fix the backticks and formatting issues.

README.md
aryaemami59: Kinda neat isn't it?

README.md
aryaemami59: Gladddddllllyyyy :)

README.md
aryaemami59: Yikes, I thought I got rid of that, guess not

README.md
aryaemami59: Yeah I will change all the examples to match the `selectX` pattern for consistency.

src/types.ts
aryaemami59: I didn't really "change" much of the types. I did added some more type expansions for hover-previews, and fixed a small inference issue with `OutputSelector`  that was coming from `ExtractReturnType`.

src/createCurriedSelectorCreator.ts
aryaemami59: I honestly think it will be helpful, it's kind of overdue actually. Would I be able give you some more confidence and perhaps change your mind by writing an insane amount of tests? It is basically a variation of `createSelector` that serves as a utility function to make our lives easier. It doesn't produce a breaking change, and I labled it as "experimental". It just seems like an appropriate API to include with a major version bump no? :)

EskiMojo14: whether it's helpful or not, it doesn't make sense to be included in a PR named "update docs for v5 changes"

README.md
markerikson: Yeah, I've used variations of that snippet a few times and definitely want it in here.

README.md
markerikson: Yep!

README.md
markerikson: Mostly, yeah.  You _could_ use it with `useSelector` to do `const {a, b} = useSelector(someStructuredSelector)`, and that's still a sort of plausible use case.  (Or you could just do `const {a, b} = useSelector(state => ({a: state.a, b: state.b}), shallowEqual), and have basically the same result.

src/createCurriedSelectorCreator.ts
markerikson: Yeah, we can discuss it after RTK 2.0 / Reselect 5.0 is out.  It's a new API, so it's something we can always release in a minor if we do decide to add it.  Right now I'm trying to stop adding or changing anything that's not strictly necessary to get RTK 2.0 out the door :)  (already punted on `createSlice.selectorFactories`, etc, so that's not holding things up.)

src/types.ts
markerikson: Yeah, can you still move those to a separate PR just so they're easier to read through?

README.md
```diff
@@ -1,707 +1,1825 @@
...
+---
 
-`createSelectorCreator` can be used to make a customized version of `createSelector`.
+## FAQ
 
-The `memoize` argument is a memoization function to replace `defaultMemoize`.
+### Why isn’t my selector recomputing when the input state changes?
 
-The `...memoizeOptions` rest parameters are zero or more configuration options to be passed to `memoizeFunc`. The selectors `resultFunc` is passed as the first argument to `memoize` and the `memoizeOptions` are passed as the second argument onwards:
+<details><summary><b>Answer</b></summary>
```
markerikson, line 1368: :hand: Actually going to remove most of these `<details>` sections from the FAQs, except for maybe a couple longer answers

markerikson:
Awright, let's get this in!

markerikson (approved):

aryaemami59:
> Awright, let's get this in!

This makes me so happy

PR #641, opened 2023-11-21
Transition README Docs to Docusaurus-Netlify Setup

aryaemami59:
This PR kickstarts migrating the existing documentation from a `README` file to a Netlify site using Docusaurus.


markerikson:
Looking at this, and I'm rather confused by the setup approach here:

- Why are we using Gulp at all for anything?  Are there other tools / approaches?
- What broke with `remark-ts-tools` in Docusaurus v3?
- Do we actually need the examples as separate TS files?

Couple nits on the `package.json` setup:

- The nested `build:example` commands are confusing
- Naming-wise, I'd prefer to have `"docs:$COMMAND"` rather than `"$COMMAND:docs"`
- Given that we use Yarn for package management, let's switch all `npm run` usages to `yarn run`

aryaemami59:
> Looking at this, and I'm rather confused by the setup approach here:
> 
> * Why are we using Gulp at all for anything?  Are there other tools / approaches?
> * What broke with `remark-ts-tools` in Docusaurus v3?
> * Do we actually need the examples as separate TS files?
> 
> Couple nits on the `package.json` setup:
> 
> * The nested `build:example` commands are confusing
> * Naming-wise, I'd prefer to have `"docs:$COMMAND"` rather than `"$COMMAND:docs"`
> * Given that we use Yarn for package management, let's switch all `npm run` usages to `yarn run`

Yeah I'm still trying to figure it all out, so a lot is going to change. To answer some of your questions:

- I'm using `gulp-preserve-typescript-whitespace` which allows me to compile TS files while preserving empty lines (currently trying to figure out a better solution).
- Not sure what broke it, but there is an issue filed for it already https://github.com/phryneas/remark-typescript-tools/issues/13
- Putting examples in separate TS files makes it easier to convert each one to markdown codeblocks, it also helps knowing each example can be run independently.
- I will fix the other issues related to `package.json`, tbh I was too sleep deprived when I wrote them :)

Another thing to note about `remark-typescript-tools` is that while it is an amazing tool and you guys use it in RTK docs, it's necessary for that project, since from the get go, the examples were written in markdown and not in TS files. So you ended up parsing markdown to TS, then TS to JS, then JS back to markdown. Which is why bulding the docs in RTK can sometimes take almost 10 minutes. If you have the examples in TS files from the beginning, you only have to do TS -> JS -> MDX instead of MDX -> TS -> JS -> MDX. That's also another reason why I set it up this way so we don't have to do what you guys did in RTK.

markerikson:
I'll admit this is sort of a weird prejudice, but I don't like Gulp :)  part of it is that it "feels" very outdated given the existence of newer tools, and I had a bad experience trying to rip it out of a project several years back.

Are there any other tool pipelines that would work here instead of Gulp?

Side note: could we also move the examples to a `docs/examples/` subfolder?

aryaemami59:
> I'll admit this is sort of a weird prejudice, but I don't like Gulp :) part of it is that it "feels" very outdated given the existence of newer tools, and I had a bad experience trying to rip it out of a project several years back.

Trust me, I know. I don't like it either. Right now I'm trying to figure out if there are other ways.

> Side note: could we also move the examples to a `docs/examples/` subfolder?

I was just thinking the same thing.

markerikson:
Okay, let's get the initial setup in so I can iterate on config

PR #644, opened 2023-11-25
suggestion: add overload for "solo selector" without input selectors

phryneas:
This is just something I wanted to explore.

Of course, people can always just call `defaultSelector`, but that's an additional function people have to be aware of, and it makes code bases less uniform. Right now, people oftentimes use identity functions, either as input selectors, or worse, even as result functions.

This would just allow a single function without input selectors to be memoized directly.

### DIFF
--- a/src/createSelectorCreator.ts
+++ b/src/createSelectorCreator.ts
@@ -141,6 +141,32 @@ export interface CreateSelectorFunction<
     OverrideArgsMemoizeFunction
   > &
     InterruptRecursion
+
+    <
+    SoloSelector extends (...args: any[]) => any,
+    OverrideMemoizeFunction extends UnknownMemoizer = MemoizeFunction,
+    OverrideArgsMemoizeFunction extends UnknownMemoizer = ArgsMemoizeFunction
+  >(
+    selector: SoloSelector,
+    createSelectorOptions?: Simplify<
+      CreateSelectorOptions<
+        MemoizeFunction,
+        ArgsMemoizeFunction,
+        OverrideMemoizeFunction,
+        OverrideArgsMemoizeFunction
+      >
+    >
+  ): OutputSelector<
+    ParametersToVirtualInputSelectors<Parameters<SoloSelector>>,
+    ReturnType<SoloSelector>,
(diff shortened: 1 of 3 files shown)


aryaemami59:
Very interseting idea, do you think it would be simpler to just have a new API called something like `createSoloSelector` or `createSimpleSelector` instead of adding another function signature to `CreateSelectorFunction`?

phryneas:
That would defeat the purpose.

People write bad code like
```js
  state => state.slice,
  slice => slice
  )
```
or
```js
  state => state,
  state => state.slice
)
```
because they don't know any other api.

They could already write
```js
const selector = defaultMemoize(state => state.slice)
```
but they don't do it, because they don't know about `defaultMemoize`.

If we now offered
```js
const selector = createSoloSelector(state => state.slice)
```
it would face the same problem - people don't know about it and won't discover it. They will resort to the bad alternatives above instead.

That's why I suggest
```js
const selector = createSelector(state => state.slice)
```
here to "just work" - it solves the problem at the root without introducing mental overhead.

aryaemami59:
That's a good point, do you think #645, will help mitigate some of the problem?

aryaemami59:
@phryneas Do you want me take a look at this to see if I can get the types to work?

phryneas:
@aryaemami59 I think #645 would help people spot the error, but currently it would still send people to use another function, which still adds the mental overload, so I think these two would complement each other.

As for fiddling with the types here: please go ahead, I don't currently have the time to take a deep dive here, and you're much more familiar with the code base. I *assume* that changing the signature order might already mitigate the warnings, but I could be wrong :)

aryaemami59:
> @aryaemami59 I think #645 would help people spot the error, but currently it would still send people to use another function, which still adds the mental overload, so I think these two would complement each other.
> 
> As for fiddling with the types here: please go ahead, I don't currently have the time to take a deep dive here, and you're much more familiar with the code base. I _assume_ that changing the signature order might already mitigate the warnings, but I could be wrong :)

I do not possess your skills, but I will certainly try my best :)

PR #645, opened 2023-11-26
Add warning in development when a result function is `x => x`.

aryaemami59:
This PR:

- Adds `noopCheck` to warn to the user if the result function returns its input. It has the same configuration options as `inputStabilityCheck`.
- Adds unit tests for `noopCheck`.
- Add documentation for `noopCheck` to `README`.
- Add documentation to docusaurus once merged with #641.
 Much like `inputStabiliyCheck` and the `noopCheck` in React-Redux, this warns the user if their result function does nothing but return its inputs.


src/utils.ts
```diff
@@ -164,6 +165,24 @@ export function runStabilityCheck(
   }
 }
 
+export const runNoopCheck = <Func extends AnyFunction>(resultFunc: Func) => {
```
EskiMojo14, line 168:
```suggestion
export const runNoopCheck = (resultFunc: AnyFunction) => {
```

src/types.ts
```diff
@@ -83,6 +83,8 @@ export interface CreateSelectorOptions<
    */
   inputStabilityCheck?: StabilityCheckFrequency
 
+  noopCheck?: StabilityCheckFrequency
```
EskiMojo14, line 86: we should probably rename the type to CheckFrequency, since this isn't a stability check

src/createSelectorCreator.ts
```diff
@@ -408,6 +418,13 @@ export function createSelectorCreator<
         arguments
       )
 
+      if (
+        process.env.NODE_ENV !== 'production' &&
+        shouldRunInputStabilityCheck(noopCheck, firstRun)
```
EskiMojo14, line 423: likewise, shouldRunCheck

EskiMojo14 (commented):
haven't had a chance to look through this properly yet, but a few initial suggestions

src/types.ts
aryaemami59: Good idea!

aryaemami59: How about `DevCheckFrequency` or `DevModeCheckFrequency` , `CheckFrequency` seems a little generic, what do you think?

aryaemami59:
BTW I only called it `noopCheck` to stay consistent with `react-redux`, if you have a better suggestion I'm all ears :)

src/types.ts
EskiMojo14: DevCheckFrequency is fine by me

EskiMojo14:
we couldn't come up with anything better for react-redux so i doubt we'll do so here :) noopCheck is fine imo

aryaemami59:
> we couldn't come up with anything better for react-redux so i doubt we'll do so here :) noopCheck is fine imo

How about `IdentityFunctionWarning` or maybe `NoTransformationWarning`?

src/types.ts
markerikson: Yep, I was literally thinking the same thing reading the current source before I got to this PR!

src/createSelectorCreator.ts
```diff
@@ -408,23 +450,28 @@ export function createSelectorCreator<
...
-        )
+          runStabilityCheck(
+            { inputSelectorResults, inputSelectorResultsCopy },
+            { memoize, memoizeOptions: finalMemoizeOptions },
+            arguments
+          )
 
-        if (firstRun) firstRun = false
+          if (firstRun) firstRun = false
+        }
```
EskiMojo14, line 474:
```suggestion
        }
        if (firstRun) firstRun = false
```

src/utils.ts
```diff
@@ -164,15 +165,43 @@ export function runStabilityCheck(
...
  * Determines if the input stability check should run.
  *
  * @param inputStabilityCheck - The frequency of the input stability check.
  * @param firstRun - Indicates whether it is the first time the selector has run.
  * @returns true if the input stability check should run, otherwise false.
  */
-export const shouldRunInputStabilityCheck = (
-  inputStabilityCheck: StabilityCheckFrequency,
+export const shouldRunDevModeCheck = (
+  inputStabilityCheck: DevModeCheckFrequency,
```
EskiMojo14, line 204: param name and jsdoc will need an update

markerikson:
I don't know how many more dev checks we'd add in the future, but I think we ought to consolidate this into a single combined setter function just in case we do add more, rather than adding a bunch of individual setters.

I'm thinking something like `setDefaultDevCheckFrequency(checks: Partial<Record<DevCheckType, DevCheckFrequency>>)`.

Let me play with that a bit on this branch.

aryaemami59:
> I don't know how many more dev checks we'd add in the future, but I think we ought to consolidate this into a single combined setter function just in case we do add more, rather than adding a bunch of individual setters.
> 
> I'm thinking something like `setDefaultDevCheckFrequency(checks: Partial<Record<DevCheckType, DevCheckFrequency>>)`.
> 
> Let me play with that a bit on this branch.

Yeah, I agree. If I put them in different files, would that be problematic?

src/createSelectorCreator.ts
EskiMojo14: currently you're only reassigning the firstRun variable inside the if block for the stability check - this means that if you have stability check disabled firstRun will always be true

markerikson:
I'm picturing having a singleton object with the current global check values, and the setter would loop over the incoming options and assign.  (Could even do `Object.assign(globalOptions, incomingOptions)` if we aren't worried about someone passing in `{stability: undefined}`, I guess.)

EskiMojo14:
the "should run" check explicitly checks against "always" and "once" so any other values will just function the same as "never", which i think is okay

src/createSelectorCreator.ts
aryaemami59: Oh wow, can't believe I that, thanks for letting me know.

aryaemami59:
@markerikson I think I get it, so it would look something like this:

```ts
export type DevModeChecks = Record<
  'inputStability' | 'identityFunction',
  DevModeCheckFrequency
>

const globalChecks: DevModeChecks = {
  inputStability: 'once',
  identityFunction: 'once'
}

export const setGlobalChecks = (devModeChecks: Partial<DevModeChecks>) => {
  Object.assign(globalChecks, devModeChecks)
}
```

then inside `createSelectorCreator`:

```ts
    const {
      memoize,
      memoizeOptions = [],
      argsMemoize = defaultMemoize,
      argsMemoizeOptions = [],
      inputStabilityCheck = globalChecks.inputStability,
      identityFunctionCheck = globalChecks.identityFunction
    } = combinedOptions
```

test/defaultMemoize.spec.ts
```diff
@@ -269,12 +269,12 @@ describe('defaultMemoize', () => {
     )
 
     fooChangeHandler(state)
-    expect(fooChangeSpy.mock.calls.length).toEqual(1)
+    expect(fooChangeSpy.mock.calls.length).toEqual(2)
```
markerikson, line 272: :question: Why are these assertions changing?

test/defaultMemoize.spec.ts
```diff
@@ -405,13 +406,7 @@ describe('defaultMemoize', () => {
     // 'a' here would _not_ recalculate
     selector('b') // ['b']
     expect(funcCalls).toBe(5)
-
-    try {
-      //@ts-expect-error issue 591
-      selector.resultFunc.clearCache()
```
markerikson, line 411: :question: what's the change about here?

README.md
```diff
@@ -1131,6 +1131,75 @@ const selectCompletedTodosLength = createSelector(
...
+
+- **Transformation Logic**: In contrast, transformation logic, such as `todos => todos.map(({ id }) => id)`, belongs in the [result function]. This is where you manipulate, format, or transform the data extracted by the input selectors.
+
+Most importantly, effective memoization in Reselect hinges on following these guidelines. Memoization, only functions correctly when extraction and transformation logic are properly segregated. By keeping extraction logic in input selectors and transformation logic in the result function, Reselect can efficiently determine when to reuse cached results and when to recompute them. This not only enhances performance but also ensures the consistency and predictability of your selectors.
+
+For memoization to work as intended, it's imperative to follow both guidelines. If either is disregarded, memoization will not function properly. Consider the following example for clarity:
+
+```ts
+// ❌ Incorrect Use Case: This will not memoize correctly, and does nothing useful!
+const brokenSelector = createSelector(
```
markerikson, line 1150: :hand: Should also show an example of creating new references in the input function, either in the same example or another example around here

yarn.lock
markerikson, line 1: Why is this lockfile change in here?

test/weakmapMemoize.spec.ts
```diff
@@ -214,3 +170,52 @@ describe('Basic selector behavior with weakMapMemoize', () => {
     expect(called).toBe(2)
   })
 })
+
+const isCoverage = process.env.COVERAGE
```
markerikson, line 174: :hand: Just for consistency, can we leave the perf test changes out of this PR?  (Also, is this duplicate of what's in #640 ?)

aryaemami59:
@markerikson I'm not too sure about the message warning message in the console, do you think we can improve on it, or leave it as is?

src/devModeChecks/inputStabilityCheck.ts
```diff
@@ -0,0 +1,47 @@
...
+  const areInputSelectorResultsEqual =
+    createAnEmptyObject.apply(null, inputSelectorResults) ===
+    createAnEmptyObject.apply(null, inputSelectorResultsCopy)
+  if (!areInputSelectorResultsEqual) {
+    // do we want to log more information about the selector?
+    console.warn(
+      'An input selector returned a different result when passed same arguments.' +
+        '\nThis means your output selector will likely run more frequently than intended.' +
+        '\nAvoid returning a new reference inside your input selector, e.g.' +
+        '\n`createSelector([(arg1, arg2) => ({ arg1, arg2 })],(arg1, arg2) => {})`',
```
markerikson, line 39: :hand: Maybe use an example with an `array.map()`?  feels like that might be more readable on that line

PR #650, opened 2023-11-30
Fix type tests for switching the default memoization method to `weakMapMemoize`

aryaemami59:
I will see if I fix the perf tests as well.


test/reselect.bench.ts
aryaemami59, line 1: This file was a duplicate of what was already in the benchmark folder.

Issue #642, opened 2023-11-25
Consider dev mode checks for `x => x` result functions


markerikson:
Just ran across these lovely functions in the wild:

- https://github.com/flipt-io/flipt/issues/2444

```ts
export const selectFlag = createSelector(
  [
    (state: RootState, namespaceKey: string, key: string) => {
      const flags = state.flags.flags[namespaceKey];
      if (flags) {
        return flags[key] || ({} as IFlag);
      }

      return {} as IFlag;
    }
  ],
  (flag) => flag
);

export const selectNamespaces = createSelector(
  [
    (state: RootState) =>
      Object.entries(state.namespaces.namespaces).map(
        ([_, value]) => value
      ) as INamespace[]
  ],
  (namespaces) => namespaces
);

export const selectCurrentNamespace = createSelector(
  [
    (state: RootState) => {
      if (state.namespaces.namespaces[state.namespaces.currentNamespace]) {
        return state.namespaces.namespaces[state.namespaces.currentNamespace];
      }
      return { key: 'default', name: 'Default', description: '' } as INamespace;
    }
  ],
  (currentNamespace) => currentNamespace
);
```

These are broken in _two_ ways:

- input selectors returning new references
- result functions that are `x => x`

We've got warnings for the first, but not for the second.

Here's another example:

```ts
export const helmChartsSelector = createSelector(
  (state: RootState) => state.main.helmChartMap,
  helmCharts => helmCharts
);
```

This isn't necessary and shouldn't even be memoized.

A really hacky thing to do would be:

- Stringify the result function
- Parse its arguments and body
- check if `args.length === 1 && args[0] === body` (conceptually. probably multiple ways to do this, ranging from actual parsing to a regex or something)

I'm actually seriously considering us adding this

Possible resources:

- https://github.com/sindresorhus/fn-args/blob/main/index.js
- https://stackoverflow.com/questions/38453881/how-to-get-an-arrow-functions-body-as-a-string
- https://stackoverflow.com/questions/6921588/is-it-possible-to-reflect-the-arguments-of-a-javascript-function
- https://stackoverflow.com/questions/3179861/javascript-get-function-body

markerikson:
Even better, Lenz just came up with this nonsense:

```ts
function checkErroneousResultFunction(resultFn){
  let wtf = false
  try {
    const obj = {}
    if (resultFn(obj) === obj) wtf = true
  } catch {}
  if (wtf) {
    throw new Error("RTFM")
  }
}
```

[...]

PR #660, opened 2023-12-10
Reduce `identityFunctionCheck` false positives

Methuselah96:
This PR attempts to reduce the number of false positives identified by `identityFunctionCheck` by making two changes:

1. Only produce a warning if the result function is only passed one argument.
2. Compare the first argument passed to the result function with the value returned by the result function. There are some edge cases where this could prevent some false positives, although admittedly, they're likely rare (one hypothetical edge case is added as a test).

### DIFF
--- a/src/createSelectorCreator.ts
+++ b/src/createSelectorCreator.ts
@@ -365,13 +365,10 @@ export function createSelectorCreator<
       )
 
       if (process.env.NODE_ENV !== 'production') {
-        const { identityFunctionCheck, inputStabilityCheck } =
-          getDevModeChecksExecutionInfo(firstRun, devModeChecks)
-        if (identityFunctionCheck.shouldRun) {
-          identityFunctionCheck.run(
-            resultFunc as Combiner<InputSelectors, Result>
-          )
-        }
+        const { inputStabilityCheck } = getDevModeChecksExecutionInfo(
+          firstRun,
+          devModeChecks
+        )
 
         if (inputStabilityCheck.shouldRun) {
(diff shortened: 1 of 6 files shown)


aryaemami59:
This seems like a good idea, the only problem is now something like this won't generate a warning:

```ts
const badSelector = createSelector(
  [(state: RootState) => state.todos, (state: RootState) => state.alerts],
  (todos, alerts) => todos
)
```

I actually thought about doing exactly what you implemented when I was initially working on it, there is definitely a trade off, not sure which one is the way to go.

Methuselah96:
Agreed, there's no way to make it perfect, but I would definitely lean towards this being a better tradeoff.

1. This still correctly warns for all practical examples given in https://github.com/reduxjs/reselect/issues/642.
2. This check has so far produced at least 8 false positives and 0 true positives in my work codebase. Each warning needs to be investigated and takes time to either individually silence, or rework (for each person who runs across them).
3. The cost of false negatives are fairly low. If the check misses a few identity functions like the example you provided, it's not that big a deal. It's unclear how often someone would write an identity function like that, whereas we do have clear practical examples for warning for a simple identity function with one arg, and for the false positives with multiple args.

All-in-all, it seems clear to me that this provides a better balance for catching true issues, and avoiding false positives for the `identityFunctionCheck`.

aryaemami59:
Yeah, currently I'm working on eslint-plugin for Reselect and RTK, and I think we can go with the solution you provided, and then the other edge cases can be caught using static code analysis. This way we can give accurate warnings without being too strict about it.

PR #681, opened 2024-01-12
feat: add codemod to convert input selectors passed as separate inline arguments to a single array

aryaemami59:
## **This PR**:

  - [X] Adds **`convertInputSelectorsToArray`** codemod to convert input selectors passed as separate inline arguments to a single array.
  
  ### Before:
  
  ```ts
  const selectTodoById = createSelector(
    (state: RootState) => state.todos,
    (state: RootState, id: number) => id,
    (todos, id) => todos.find((todo) => todo.id === id)
)
  ```
  
  ### After:
  
  ```ts
  const selectTodoById = createSelector(
    [(state: RootState) => state.todos, (state: RootState, id: number) => id],
    (todos, id) => todos.find((todo) => todo.id === id)
)
  ```

  - [X] This should also help with transitioning into using the **`withTypes`** API since currently it does not support input selectors passed as separate inline arguments.
  - [X] Tested with **`yalc`** to make sure the CLI itself works.
  - [x] Added proper documentation for this codemod.


markerikson:
Nits: 

- can we move the codemod tests lower in the file so that the TS and unit test jobs show up first?
- I don't think we need to test on both Ubuntu and MacOS. Ubuntu seems sufficient.

PR #683, opened 2024-01-13
Remove useCallback example from weakMapMemoize page

### LINKED ISSUE #682: Incorrect weakMapMemoize alternative example using useCallback
The example of avoiding redundant selector calls by using useCallback on the [weakMapMemoize page](https://reselect.js.org/api/weakMapMemoize) is non-working
We still have the same instance of selector, so the cache will share to all the component instances using the same selector

https://codesandbox.io/p/sandbox/reselect-usecallback-demo-y8wldt

benax-se:
Remove wrong selector with useCallback usage example at the weakMapMemoize page

Closes #682


EskiMojo14:
thanks!

Issue #694, opened 2024-02-02
Library do not work in Safari < 14.1


veksa:
Error in console when including the library:

`Unexpected token '=' . Expected an opening '(' before a method's parameter list`

The same problem is described here: https://github.com/chakra-ui/chakra-ui/issues/6828

Setting the target to `es2019` in the esm build will fix the problem for High Sierra to Catalina.

PR: https://github.com/reduxjs/reselect/pull/695

markerikson:
We specifically target ES2020 out of the box for the release.

If you need to use older browsers, either transpile the default "modern" build artifact yourself, or use the "legacy" artifact already included in the package (which targets ES2017).

PR #695, opened 2024-02-02
Set target to es2019 on esm build

veksa:
Library do not work in Safari < 14.1

`Unexpected token '=' . Expected an opening '(' before a method's parameter list`

Setting the target to es2019 in the esm build will fix the problem for High Sierra to Catalina.


markerikson:
Per #694, we explicitly target modern JS syntax by default - either transpile the dependency as part of your build step, or alias to use the "legacy" build artifact in the package.

PR #698, opened 2024-02-11
Ensure `lruMemoize` correctly memoizes when `maxSize` is set to a number less than 1

aryaemami59:
This PR:

  - [X] Fixes an issue with `lruMemoize` that would cause memoization to completely break when `maxSize` is set to a value that is less than 1.
  - [X] Adds units tests to make sure the issue is effectively resolved.


EskiMojo14:
to be honest, i'm not sure this is something we needed to account for, but ultimately it's a one character change so can't hurt.

thanks!

Issue #693, opened 2024-01-31
`weakMapMemoize` with `resultEqualityCheck` is provided empty objects for first call.


keegan-lillo:
Hi! Thanks for all the great QOL improvements in Reselect 5, however I've found a pretty major issue with the `resultEqualityCheck`  option when using the `weakMapMemoize` memoization. It calls the `resultEqualityCheck` even if it is the first time the selector has run. This wouldn't be too bad, however this first call is called with empty objects instead of the actual result of the selector. Please see the example below. Obviously this is a bit of a contrived example using strings, but things get pretty hairy if you expect the result to be a `Set` or another complex object.

```ts
const firstMessageSelector = createSelector(
  [(messages) => messages[0]],
  (message) => `${message}!!!`,
  {
    memoizeOptions: {
      resultEqualityCheck: (a, b) => {
        console.log("resultEqualityCheck:", { a, b });
        return a === b;
      },
    },
  }
);

const one = firstMessageSelector(["one", "two"]);
console.log("one:", one);
const two = firstMessageSelector(["two", "three"]);
console.log("two:", two);
```

Expected Console Output:
```
firstMessageSelector: one
one: one!!!
firstMessageSelector: two 
resultEqualityCheck: Object { a: "one!!!", b: "two!!!" }
two: two!!! 
```

Actual Console Output:
```
firstMessageSelector: one
resultEqualityCheck: Object { a: {}, b: {} }
one: one!!!
firstMessageSelector: two 
resultEqualityCheck: Object { a: "one!!!", b: "two!!!" }
two: two!!! 
```

EskiMojo14:
hmm, this is a good point - the `resultEqualityCheck` is typed to only expect the result of the memoized function, but the [`inputStabilityCheck`](https://github.com/reduxjs/reselect/blob/master/src/devModeChecks/inputStabilityCheck.ts#L33) uses the same memoizeOptions and memoize with a *different* function, one that always returns a new empty object. This is important for how the check works.

I'm not sure of the best fix for this, but I agree it's not great.

braeden:
Really excited about `reselect` v5 but +1 on this is upgrade blocking 😞 . We almost guarantee a (development-only) runtime break when `resultEqualityCheck` does any sort of complex work. (and disabling `inputStabilityCheck` feels like a band-aid fix)

----

I could be missing something @EskiMojo14, but this is the code I would have expected for `inputStabilityCheck`:

```ts
[...]

PR #699, opened 2024-02-11
Fix `resultEqualityCheck` behavior in `weakMapMemoize`

aryaemami59:
## Overview
Previously using [`weakMapMemoize`](https://reselect.js.org/api/weakMapMemoize)with `resultEqualityCheck` would cause the `resultEqualityCheck` to be called with empty objects as arguments due to how [`inputStabilityCheck`](https://reselect.js.org/api/development-only-stability-checks#inputstabilitycheck) works. Here's a summary of the changes:

**This PR**:

  - [X] Fixes `resultEqualityCheck` in [`weakMapMemoize`](https://reselect.js.org/api/weakMapMemoize) so that when input selectors are stable, `resultEqualityCheck` does not get called on the first run of the output selector.
  - [X] Adds runtime unit tests to verify that:
    - `resultEqualityCheck` works correctly when set to `shallowEqual`.
    - `resultEqualityCheck` is not called on the first output selector call.
    - [`weakMapMemoize`](https://reselect.js.org/api/weakMapMemoize)/[`lruMemoize`](https://reselect.js.org/api/lruMemoize) with `resultEqualityCheck` set to `referenceEqualityCheck` works the same as [`weakMapMemoize`](https://reselect.js.org/api/weakMapMemoize)/[`lruMemoize`](https://reselect.js.org/api/lruMemoize) without `resultEqualityCheck`.
  - [X] Adds benchmarks to measure the `resultEqualityCheck` performance impact.
  - [X] Adds unit tests to test the dynamics between [`inputStabilityCheck`](https://reselect.js.org/api/development-only-stability-checks#inputstabilitycheck) and `resultEqualityCheck`.
  - [X] Fixes a specific bug within [`weakMapMemoize`](https://reselect.js.org/api/weakMapMemoize) where setting `resultEqualityCheck` to `referenceEqualityCheck` would prevent `memoizedResultFunc.resultsCount()` from incrementing upon result recalculation.
  - [X] Fixes a performance issue in [`weakMapMemoize`](https://reselect.js.org/api/weakMapMemoize) when comparing two memoized functions, one with `resultEqualityCheck` set to `referenceEqualityCheck` and the other without `resultEqualityCheck`. Previously, there was a noticeable discrepancy between the two, which has now been corrected to achieve the desired effect.

### DIFF
--- a/src/weakMapMemoize.ts
+++ b/src/weakMapMemoize.ts
@@ -227,25 +227,29 @@ export function weakMapMemoize<Func extends AnyFunction>(
       // Allow errors to propagate
       result = func.apply(null, arguments as unknown as any[])
       resultsCount++
+
+      if (resultEqualityCheck) {
+        const lastResultValue = lastResult?.deref?.() ?? lastResult
+
+        if (
+          lastResultValue != null &&
+          resultEqualityCheck(lastResultValue as ReturnType<Func>, result)
+        ) {
+          result = lastResultValue
+
+          resultsCount !== 0 && resultsCount--
+        }
+
+        const needsWeakRef =
+          (typeof result === 'object' && result !== null) ||
+          typeof result === 'function'
+
+        lastResult = needsWeakRef ? new Ref(result) : result
+      }
     }
 
     terminatedNode.s = TERMINATED
 
-    if (resultEqualityCheck) {
-      const lastResultValue = lastResult?.deref?.() ?? lastResult
-      if (
-        lastResultValue != null &&
-        resultEqualityCheck(lastResultValue as ReturnType<Func>, result)
-      ) {
-        result = lastResultValue
-        resultsCount !== 0 && resultsCount--
-      }
-
-      const needsWeakRef =
-        (typeof result === 'object' && result !== null) ||
-        typeof result === 'function'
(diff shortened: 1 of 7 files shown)


keegan-lillo:
Thanks for working on this! I believe this PR is related to this issue: https://github.com/reduxjs/reselect/issues/693

EskiMojo14:
related, but not a complete fix - if inputs aren't stable, your resultEqualityCheck will still be called with an empty object

aryaemami59:
@keegan-lillo This should mitigate the problem as long as your input selectors are stable. It will at least make the default configs a tiny bit less problematic. But @EskiMojo14 is correct, it doesn't fully address the issue.

EskiMojo14:
this looks good, but has conflicts - happy to merge once fixed

EskiMojo14 (approved):

EskiMojo14:
thanks!

braeden:
Hey @EskiMojo14, any chance a release could be cut which contains this change?

markerikson:
Sorry for the delay on this. I should have time to get it out on Saturday.

markerikson:
Okay, out in https://github.com/reduxjs/reselect/releases/tag/v5.1.1 !

PR #701, opened 2024-02-19
LRU implementation with hashmap and doubly linked list

kuldeepsinghborana:


EskiMojo14:
this doesn't seem to use the equality function?

kuldeepsinghborana:
> this doesn't seem to use the equality function?

@EskiMojo14 Yes, the equality function can not be used with a hashmap, we need to iterate throgh the all values which will not give us any advantage our the existing memoiser

markerikson:
@kuldeepsinghborana yeah, that's a problem, and that's what i was pointing to in the issue thread. We rely on equality checks, not a single key.

kuldeepsinghborana:
@markerikson  @EskiMojo14  @aryaemami59 
Can we create a new memoizer that takes a keyGen (key generator) function as an argument? The default keyGen function would convert the supplied argument into a JSON string. Subsequently, our cache Map would utilize this string as a key.

To enhance predictability in the JSON string, we can implement key sorting, ensuring that keys are consistently ordered for every object.

aryaemami59:
@kuldeepsinghborana I actually did this very recently when I was experimenting with creating a new memoizer, @markerikson is right, `JSON.stringify` is very expensive, however you can do `array.join()` to stringify the arguments instead.

markerikson:
`array.join()` isn't going to help either.  `{a: 1}` and `{a: 1}` are both going to stringify to `"[object Object]"`, which is not unique.

PR #706, opened 2024-02-21
Add `.gitattributes` file to make line endings more consistent

aryaemami59:
**This PR**:

  - [X] Adds `.gitattributes` file to make line endings more consistent.


timdorr:
This is silly, we should have merged this in already. Thanks!

PR #748, opened 2025-06-12
perf(lruMemoize): avoid reallocating entries

romgrk:
Avoid allocating a new entries array on each call.


markerikson:
Ended up going a different direction in #779 - thanks for the inspiration!

PR #759, opened 2026-06-23
fix(`lruMemoize`): pass (cached, new) argument order to `equalityCheck` in LRU cache

JSap0914:
## Summary

`createSingletonCache` (used when `maxSize <= 1`) and `createLruCache` (used when `maxSize > 1`) call the key-comparator with **reversed argument orders**:

| Implementation | Call | `equalityCheck` receives |
|---|---|---|
| `createSingletonCache.get` | `equals(entry.key, key)` | `(cached, new)` ✓ |
| `createLruCache.get` | `equals(key, entry.key)` | `(new, cached)` ✗ |

`createCacheKeyComparator` names its parameters `(prev, next)` and calls `equalityCheck(prev[i], next[i])`, making the expected contract clear: **cached/prev first, new/next second**. The singleton cache honours this; the LRU cache inverts it.

For the default symmetric `referenceEqualityCheck (a, b) => a === b` there is no visible difference. But any custom **asymmetric** `equalityCheck` will behave differently depending on whether `maxSize` is 1 or greater than 1, which is a correctness bug.

## Fix

One-character swap in `createLruCache.get`:

```diff
-    const cacheIndex = entries.findIndex(entry => equals(key, entry.key))
+    const cacheIndex = entries.findIndex(entry => equals(entry.key, key))
```

## Tests

Adds a regression test that records every `(a, b)` pair passed to a custom `equalityCheck` and asserts the cached value always comes first—for both `maxSize=1` (singleton) and `maxSize=2` (LRU). The test fails on `master` and passes after the fix. All 18 pre-existing tests continue to pass.

> This PR was developed with AI assistance.

### DIFF
--- a/src/lruMemoize.ts
+++ b/src/lruMemoize.ts
@@ -52,7 +52,7 @@ function createLruCache(maxSize: number, equals: EqualityFn): Cache {
   let entries: Entry[] = []
 
   function get(key: unknown) {
-    const cacheIndex = entries.findIndex(entry => equals(key, entry.key))
+    const cacheIndex = entries.findIndex(entry => equals(entry.key, key))
 
     // We found a cached entry
     if (cacheIndex > -1) {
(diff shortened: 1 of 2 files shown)


markerikson:
Superseded by #777 , thanks!

PR #761, opened 2026-07-19
feat: add `maxSize` option to `weakMapMemoize` to bound cache growth (#635)

### LINKED ISSUE #635: Investigate potential memory leak issues with `weakmapMemoize`
Copying notes from @phryneas : 

> **phryneas**: Thoughts on combining that with a max cache size like the weak cache I showed you?
> **eskimojo**: is having a maximum cache size desirable?
> **phryneas**: If your cache approaches 5000 elements that will never be read from again, very much so
> **phryneas**: I've been doing a lot drilling into memory leaks recently and a lot of them are weakMaps with keys that hang around for too long: 
> **phryneas**: Eskimojo - so you know what PR we're talking about: https://github.com/benjamn/optimism/pull/599
> **phryneas**: : Yeah.. looking at `WeakMapMemoize` it's very close to https://www.npmjs.com/package/@wry/trie and a lot of the reasons for Apollo Client memory leaks.
> It's funny, because that way reselect essentially becomes https://www.npmjs.com/package/optimism 😅 
> 
> So here's what I'd do: use the current `weakMapMemoize` implementation as a way of generating unique cache keys, but never store any actual data in there. Instead, use those keys as cache keys to a "flat" cache with a maximum size - if possible weakly indexed, like the implementation I linked above.
> **phryneas**: Right now, the saving grace is that your root selector will always be an object (the state), and hopefully at some point all of it's references will disappear, which will prune a full Trie.
But that's not a guarantee anymore for child selectors (or selectors that are just used with something different than a state). Imagine they are called with (primitive, primitive, primitive). Every result for every possible primitive combination that was ever input into one of those selectors will be kept around for the full application lifetime.
> **phryneas**: And at least for the second plan (or the first plan, assuming something keeps state values around for longer) you need some kind of max cache size - at least if you make it the default and people don't have to carfully think about their choices

veksa:
# feat: add `maxSize` option to `weakMapMemoize` to bound cache growth

Fixes #635

## Problem

`weakMapMemoize` builds a tree of cache nodes keyed by the identity of each
argument. Object/function arguments are stored in `WeakMap`s and are cleaned up
by the garbage collector once the argument becomes unreachable. Primitive
arguments, however, are stored in regular strong `Map`s.

When a selector is called with a long-lived object as the first argument (for
example the Redux state, or a stable value derived by a sub-selector) followed
by ever-changing primitives, every result that was ever computed is retained for
the entire lifetime of that object. Nothing is evicted, so the strong `Map`s
grow without bound — a de-facto memory leak.

```ts
const selector = weakMapMemoize(
  (array: number[], from: number, to: number) => array.slice(from, to)
)

const state = Array.from({ length: 10_000 }, (_, i) => i)
const first = selector(state, 2000, 2500)

for (let i = 0; i < 10_000; i++) selector(state, i, i + 100)

selector(state, 2000, 2500) === first // true — still cached, forever
```

A realistic trigger is an endlessly scrolling list backed by a selector such as
`(state, from, to) => state.items.slice(from, to)`: as `from`/`to` change with
every scroll event, every intermediate slice stays in memory for as long as the
state object is alive.

## Change

Add an opt-in `maxSize` option to `WeakMapMemoizeOptions`:

```ts
weakMapMemoize(fn, { maxSize: 10 })
```

- Caps the number of cached results using a least-recently-used (LRU) policy.
- When the limit is exceeded, the least-recently-used result is evicted and any
  now-empty ancestor `Map` branches are pruned, so the strong primitive `Map`s
  stay bounded and evicted results become eligible for garbage collection.
- Object (`WeakMap`) branches continue to rely on normal garbage collection and
  are not pruned.

### Design notes

- **Backwards compatible and zero-cost when unused.** Without `maxSize` the cache
  is unbounded, exactly as before. The LRU bookkeeping and the extra node
  back-references (`parent`/`map`/`key`) are only maintained when `maxSize` is
  set, so the hot path for existing users is unchanged.
- The default behavior is intentionally left unbounded. Making `weakMapMemoize`
  bounded by default is a separate, breaking decision and is out of scope here.
- `maxSize` is validated to be a positive integer.
- `clearCache()` resets the LRU state.

### Bounding a `createSelector`

`createSelector` memoizes on two levels: `argsMemoize` caches on the raw
selector arguments, and `memoize` caches the result function on the extracted
input values. Both hold results keyed by primitive arguments, so both must be
bounded — otherwise `argsMemoize` short-circuits and the leak remains:

```ts
const selectSlice = createSelector(
  [(state: RootState) => state.items, (_: RootState, from: number) => from],
  (items, from) => items.slice(from),
  {
    memoize: weakMapMemoize,
    memoizeOptions: { maxSize: 10 },
    argsMemoize: weakMapMemoize,
    argsMemoizeOptions: { maxSize: 10 }
  }
)
```

## Tests

`test/weakMapMemoize.memory.spec.ts` adds a `toBeGarbageCollected` matcher
(garbage collection is exposed via `node --expose-gc`, already used by the
`test` script) and covers:

- The unbounded behavior on the current implementation: an old result stays
  cached after many distinct calls and is not collected while the state object
  remains reachable.
- `maxSize` validation.
- Caching within the configured size, LRU eviction once it is exceeded, and a
  recently-used result surviving while a staler one is evicted.
- Eviction of object-keyed results (the `WeakMap` branch), `clearCache()`
  resetting the bounded cache, and coexistence with `resultEqualityCheck`.
- Bounding a full `createSelector` via `maxSize` on both `memoizeOptions` and
  `argsMemoizeOptions`.
- With a small `maxSize` and many distinct primitive combinations, an early
  result is garbage collected while the state object and the selector stay
  alive.

### DIFF
--- a/.github/workflows/build-and-test-types.yml
+++ b/.github/workflows/build-and-test-types.yml
@@ -125,8 +125,10 @@ jobs:
           path: .
 
       # Note: We currently expect "FalseCJS" failures for Node16 + `moduleResolution: "node16",
+      # Pinned to 0.18.5+: 0.18.2 crashes ("Cannot read properties of undefined (reading 'filename')")
+      # because it kept only the last (empty) gunzip chunk; fixed by concatenating all chunks.
       - name: Run are-the-types-wrong
-        run: npx @arethetypeswrong/cli@0.18.3 ./package.tgz --format table --ignore-rules false-cjs
+        run: npx @arethetypeswrong/cli@0.18.5 ./package.tgz --format table --ignore-rules false-cjs
 
   test-published-artifact:
     name: Test Published Artifact ${{ matrix.example }}
--- a/src/weakMapMemoize.ts
+++ b/src/weakMapMemoize.ts
@@ -31,34 +31,13 @@ const Ref = /* @__PURE__ */ getWeakRef()
 const UNTERMINATED = 0
 const TERMINATED = 1
 
-interface UnterminatedCacheNode<T> {
-  /**
-   * Status, represents whether the cached computation returned a value or threw an error.
-   */
-  s: 0
-  /**
-   * Value, either the cached result or an error, depending on status.
-   */
-  v: void
-  /**
-   * Object cache, a `WeakMap` where non-primitive arguments are stored.
-   */
-  o: null | WeakMap<Function | Object, CacheNode<T>>
-  /**
-   * Primitive cache, a regular Map where primitive arguments are stored.
-   */
-  p: null | Map<string | number | null | void | symbol | boolean, CacheNode<T>>
-}
-
-interface TerminatedCacheNode<T> {
-  /**
-   * Status, represents whether the cached computation returned a value or threw an error.
-   */
-  s: 1
-  /**
-   * Value, either the cached result or an error, depending on status.
-   */
-  v: T
+/**
+ * Fields shared by every cache node, regardless of termination status. The
+ * back-references (`parent`/`map`/`key`) and LRU links (`lru`/`mru`) are only
+ * used when a `maxSize` is configured, so that terminated nodes can be evicted
+ * and their now-empty ancestors pruned out of the strong primitive `Map`s.
+ */
+interface CacheNodeBase<T> {
   /**
    * Object cache, a `WeakMap` where non-primitive arguments are stored.
    */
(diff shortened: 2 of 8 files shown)


markerikson:
Ended up going with a very different implementation (generational architecture) in #783 , but thanks for the inspiration and pushing to make this happen!

PR #762, opened 2026-07-24
Clarify that `weakMapMemoize` is the default memoization function (v5)

### LINKED ISSUE #753: Documentation unclear that `weakMapMemoize` is now the default memoizer
as a developer i found mention about true behavior of memoization here https://reselect.js.org/introduction/v5-summary#breaking-changes 
But i found this page only after playing with reselect in sandboxes. I was confused why  this example may no effect: https://reselect.js.org/api/weakmapmemoize/#examples

llm generated: 
## Problem

The documentation doesn't clearly state that `weakMapMemoize` is now the default memoization function for `createSelector` (changed in commit [1ffd434](https://github.com/reduxjs/reselect/commit/1ffd434ba6c956f7d43ad8081d3042b530f0e228)).

This causes confusion because:

1. **Examples show redundant configuration**: The [weakMapMemoize example](https://github.com/reduxjs/reselect/blob/7d7b33610e5f2ba7c46f89cb7bc0358e365cc070/docs/examples/weakMapMemoize/cacheSizeSolution.ts#L4) and [API documentation](https://reselect.js.org/api/weakMapMemoize) show:

```ts
const selectItemsByCategory = createSelector(
  [
    (state: RootState) => state.items,
    (state: RootState, category: string) => category
  ],
  (items, category) => items.filter(item => item.category === category),
  {
    memoize: weakMapMemoize,      // ← This is already the default
    argsMemoize: weakMapMemoize   // ← This is also already the default
  }
)
```

2. **Developers may not realize the default changed**: Those familiar with Reselect v4.x (where `lruMemoize`/`defaultMemoize` was default) may not realize the behavior has changed.

3. **Migration implications unclear**: The difference between `lruMemoize` (cache size: 1) and `weakMapMemoize` (unlimited cache) has significant performance and memory implications that aren't highlighted.

## Suggested Improvements

1. **Add a callout to the main documentation** stating:
   > ⚠️ **Note**: As of v5.0, `weakMapMemoize` is the default memoizer. Previous versions used `lruMemoize` with a cache size of 1.

2. **Update examples** to either:
   - Remove the explicit `memoize`/`argsMemoize` options (since they're redundant)
   - Add comments explaining that this is showing explicit configuration but isn't necessary:
   ```ts
   {
     // These are the defaults in v5.0+, shown here for clarity:
     memoize: weakMapMemoize,
     argsMemoize: weakMapMemoize
   }
   ```

## Current Versions Affected
- v5.x documentation

veksa:
Closes #753

I was reading through the API docs and hit the same confusion the issue describes: nowhere does it actually say that `createSelector` switched to `weakMapMemoize` by default in v5. You kind of have to already know it.

A couple of things made it worse. The `createSelector` examples pass `memoize: weakMapMemoize` and `argsMemoize: weakMapMemoize` explicitly, so it looks like you're supposed to configure that yourself — even though it's just the default now. And if you're coming from v4 (where the default was `lruMemoize` with a cache size of 1), there's no signal that the caching behavior changed under you.

While digging in I also noticed two pages that flat-out state the wrong default — `createSelectorCreator` and `createStructuredSelector` both say the args memoizer defaults to `lruMemoize`, which contradicts the source and the v5 summary.

So this PR:

- Adds a short note at the top of `createSelector` explaining that `memoize` and `argsMemoize` both default to `weakMapMemoize` since v5, how that differs from the old `lruMemoize` cache-of-1 behavior, why it helps (effectively unbounded, identity-keyed cache — no more "cache size of 1" surprise when a selector is called with many different args), the memory tradeoff (entries live as long as their args are referenced, then get GC'd), and that you don't need to pass these options unless you want to override them.
- Adds a note on the `weakMapMemoize` page saying it's the default now, and that the examples there pass `memoize`/`argsMemoize` explicitly only to show the behavior — a plain `createSelector(...)` already uses it.
- Fixes the `lruMemoize` → `weakMapMemoize` default in the `createSelectorCreator` and `createStructuredSelector` type-parameter tables.

Docs only. I double-checked the real defaults against `src/createSelectorCreator.ts` (`argsMemoize` defaults to `weakMapMemoize`, and the exported `createSelector` is built with it). No example source files changed, so nothing needs regenerating, and everything passes Prettier.


markerikson:
Added in #783 , thanks!

PR #765, opened 2026-07-26
Add tsgo (TypeScript 7 native preview) type-check to CI

veksa:
## Summary

TypeScript 7 ships as a native (Go) compiler, distributed today as `@typescript/native-preview` with a `tsgo` binary. This PR adds a CI job that type-checks Reselect's type tests with `tsgo`, and makes the small set of changes needed for the whole `typescript_test` suite to pass cleanly under both `tsc` and `tsgo`. Everything stays green on the existing `tsc` matrix (5.0 through 5.8), so the day TS 7 becomes the compiler we build with, there's nothing left to fix.

## Why

`tsgo` is close enough that it's worth running in CI today rather than discovering a wall of incompatibilities later. Pointing it at a heavily overload-driven library like Reselect turned out to be a good stress test, and it surfaced a few genuine forward-compatibility items in our config plus one interesting difference in how errors are reported.

## What's in here

**Config that TypeScript 7 no longer accepts.** `tsgo` rejects a few options that were removed in TS 7, so `tsconfig.json` needed updating: `moduleResolution: "Node"` (node10) becomes `"bundler"`, the now-removed `baseUrl` is dropped, and the `paths` entries get the leading `./` that TS 7 requires. These are no-ops for `tsc` 5.x and for the Vitest type tests, which still pass unchanged.

**Type-test lib and globals.** The `typescript_test` project pinned `lib` to just `ES2021.WeakRef`. Classic `tsc` pulls the rest of the standard library in through that file's reference chain, but `tsgo` doesn't, so it's now `ES2021` (a superset — same APIs). It also picks up `types: ["node"]` because `tsgo` doesn't auto-include `@types/node` globals the way `tsc` does, and a couple of `src` files touch `console`/`process`.

**The negative type tests.** This is the bulk of the diff, and it's mostly deletions. Our `@ts-expect-error` assertions for the "this call should be a type error" cases spanned multi-line calls, with the directive sitting on the line `tsc` happens to report the overload error on. `tsgo` reports the same errors — it just pins them to the specific offending argument instead — which left those directives looking unused. Since `@ts-expect-error` is tied to the exact next line, there's no multi-line arrangement that satisfies both compilers, so each of these calls is now collapsed onto a single line with one leading `@ts-expect-error`. The assertions stay exactly as strict; they simply stop caring about where the squiggle is drawn. Positive tests and everything else are untouched.

**One relaxed assertion.** `createStructuredSelector(...).dependencies` is currently inferred in a different tuple order by `tsgo` than by `tsc`. That single spot swaps its exact-tuple `expectExactType` for a membership check on the same three signatures, with a comment explaining why. The runtime ordering is still covered by the regular unit tests.

**The wiring.** `@typescript/native-preview` is added as a dev dependency, there's a `type-check:tsgo` script (`tsgo --noEmit -p typescript_test/tsconfig.json`, mirroring `test:typescript`), and a new `test-types-tsgo` job in the CI workflow that runs it.

## Notes

The dev build is pinned rather than tracking `latest` so the job doesn't turn red on an unrelated upstream change. When TS 7 stabilises this can move to a real release and, if we ever want it, into the version matrix alongside the `tsc` runs.


EskiMojo14:
isn't typescript 7 in stable release now, so we don't need the preview package? it should be enough just to add 7 to the matrix

we also follow the [DefinitelyTyped window](https://github.com/definitelytyped/definitelytyped#support-window) so you can remove anything older than 5.6

veksa:
@EskiMojo14 

Good catch, TS 7 is stable now so the preview package was overkill. Dropped the native-preview dep and the separate tsgo job — just added 7.0 (and 6.0) to the type matrix so it runs with the normal tsc like everything else. Also trimmed the versions below 5.6 to match the DefinitelyTyped window. Checked locally that the types pass on both 6.0.3 and 7.0.2.

EskiMojo14:
i'll merge #764 once CI passes on that, so this can then be rebased and we can get a better CI picture

markerikson:

PR #767, opened 2026-07-27
Migrate from ESLint to oxlint

veksa:
This swaps ESLint out for [oxlint](https://oxc.rs/docs/guide/usage/linter). The main motivation is the upcoming move to TypeScript 7 (typescript-go): the `@typescript-eslint` toolchain is tied to the TypeScript compiler API and becomes a blocker there, whereas oxlint is a standalone Rust linter that doesn't depend on it. It's also dramatically faster and drops a large chunk of the dependency tree — installing this branch removes around 150 transitive packages.

## What changed

The old `.eslintrc` and `.eslintignore` are gone, replaced by a single `.oxlintrc.json`. It turns on the `correctness` category as errors and keeps the two rules the project actually relied on: `consistent-type-imports` (with `separate-type-imports`, same as before) and the relaxations for test and example files. The `lint` script now runs `oxlint src test`, and all the `@typescript-eslint` / `eslint` / `eslint-plugin-*` dev dependencies are replaced by a single `oxlint` entry. The few inline `eslint-disable` comments were rewritten to the `oxlint-disable` equivalents.

## Code fixes

oxlint's `correctness` category is stricter than `eslint:recommended`, so the migration surfaced 44 findings. I fixed all of them in place rather than silencing rules:

- In `src`: replaced the `Object` wrapper type with `object` in `weakMapMemoize`, dropped the redundant parameter-property assignments in `proxy`, and turned the `x !== 0 && x--` short-circuits into plain `if` statements.
- In `test`: the Chai getter-style assertions (`.that.is.not.empty`, `.to.be.true`) were rewritten to their callable forms so they no longer look like unused expressions, without re-evaluating any selector (so recomputation counts stay intact). The bare property reads in the benchmarks now feed a sink variable, and a couple of dead expression statements were tidied up.

## Follow-up: `consistent-type-exports` needs TypeScript 7

The old config also had `@typescript-eslint/consistent-type-exports` enabled. In oxlint that rule is only available through type-aware linting, which requires TypeScript 7 (typescript-go) and the `oxlint-tsgolint` package. I left it out for now because typescript-go rejects the current tsconfigs — `moduleResolution: "Node"` (node10) and `baseUrl` have been removed in TS7, and the `paths` entries need to be relative. There's a `TODO(ts7)` in `.oxlintrc.json` describing exactly what to add once the tsconfigs are migrated, so restoring the rule is a small follow-up on top of the TS7 work.


markerikson:

PR #768, opened 2026-07-27
Migrate from Prettier to oxfmt

veksa:
This replaces Prettier with [oxfmt](https://oxc.rs/docs/guide/usage/formatter), the formatter from the Oxc project. oxfmt is Prettier-compatible but written in Rust, so it's dramatically faster and lets us drop Prettier from the dependency tree.

## What changed

The config was generated with oxfmt's built-in `--migrate=prettier`, so `.oxfmtrc.json` carries over every setting from the old `.prettierrc.json` one to one — no semicolons, single quotes, two-space indent, no trailing commas, avoid arrow parens. The migration also picked up that `printWidth` wasn't set and pinned it to Prettier's default of 80 (oxfmt's own default is 100), which keeps the wrapping identical instead of reflowing the whole codebase. The old `.prettierrc.json` is gone, `prettier` is swapped for `oxfmt` in devDependencies, and the `format` script now runs `oxfmt src test`. There was no `.prettierignore`, and oxfmt reads `.gitignore` on its own, so nothing was lost there.

I also added a `format:check` script and wired it into CI right after the lint step. Formatting was never actually enforced before — Prettier only ran locally in `--write` mode — so this closes that gap and stops unformatted code from slipping in.

## Reformatting

oxfmt is close enough to the current Prettier output that only four files in `src` change, and those are the usual Prettier v2 to v3 differences in how nested ternaries and broken-up call arguments are indented (the repo was still on `prettier@2`). Running the formatter is idempotent afterwards.

One spot needed a manual touch: in `src/types.ts` there were two line comments sitting between `extends` and the function type, and oxfmt collapsed them onto one line and flipped their order, which turned them into nonsense. I moved the comment inside the parentheses as a leading comment on the parameter — a position the formatter handles cleanly — without changing the type itself.

### DIFF
--- a/.github/workflows/build-and-test-types.yml
+++ b/.github/workflows/build-and-test-types.yml
@@ -38,6 +38,9 @@ jobs:
       - name: Run linter
         run: yarn lint
 
+      - name: Check formatting
+        run: yarn format:check
+
       - name: Run tests
         run: yarn test
 
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -1,0 +1,11 @@
+{
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
+  "semi": false,
+  "singleQuote": true,
+  "tabWidth": 2,
+  "trailingComma": "none",
+  "arrowParens": "avoid",
+  "printWidth": 80,
+  "sortPackageJson": false,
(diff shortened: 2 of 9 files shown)


markerikson:

